PR #1, opened 2017-06-29
Add a simple check to see if plain http links are found

jgkamat:
Will fail if `http[^s]` is found. Let me know if you think this is wrong. 

This isn't a multiple status token, but it should be pretty clear in the circle ui.

We can make it a multiple status token once circleci 2.0 workflows support forked PR's: https://circleci.com/docs/2.0/faq/#can-i-build-fork-prs-using-workflows


jgkamat:
eh yaml is hard, will fix in a bit

kberzinch (approved):

PR #2, opened 2017-07-01
Add CI check for linear slides

kberzinch:
Based on discussion with other presenters, we'd like to stick to linear slides.

(Am I doing it right?)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,9 +14,6 @@ jobs:
       - run:
           name: No HTTP Check
           command: sh -c "! grep -r 'http[^s]' src/"
-      - run:
-          name: Linear Slides Only
-          command: sh -c "! grep -r "^[*][*]" src/"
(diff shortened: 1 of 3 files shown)


jgkamat:
Setting `reveal-hlevel`  to `0` should give you the effect you want.

kberzinch:
Neat! Will update.

jgkamat (approved):
LGTM :+1:

jgkamat:
Also keep in mind, some slides [like this one, for example](https://raw.githubusercontent.com/RoboJackets/software-training/weekOneGitSlides/src/github.org) will continue to have nested slides because they override the project-wide setting in the file itself. It's probably a good idea to remove those headers if you don't explicily want to override project settings (although, for that example, I think that's a WIP by @chsahit).

kberzinch:
Yeah, I think we can catch those in PR review.

PR #4, opened 2017-07-03
Week one topic slides

kberzinch:
Still needs a few tweaks.

 - [x] Each presenter needs an introduction slide.
 - [x] We may want to discuss the purpose of RJ software training in a bit more detail - I wrote up a couple bullet points but I'm not sure if that covers everything.
 - [x] My Linux terminal summary needs help - I think it covers everything important but I don't like some of the wording.


src/week1.org
```diff
@@ -0,0 +1,41 @@
...
+- (Almost) everything is a file
+- There's one big happy file system
+- Lots of tiny tools you can chain together to do useful things
+
+* Basic commands
+| Command | Use |
+|-------+------|
+| =cd= | Change to a different directory |
+| =ls=  | List files in this directory |
+| =mkdir= | Make a new directory |
```
jgkamat, line 126: Most people won't know what a directory is. Maybe say `folder/directory`

src/week1.org
```diff
@@ -0,0 +1,41 @@
...
+| =rm= | Remove a file |
+| =rmdir= | Remove a directory |
```
jgkamat, line 128: I find this is not too useful unless you want to remove empty directories. Maybe we can replace it with `rm -r`? Not sure if that's a good idea though 😄

src/week1.org
jgkamat: You might also want to make a "terminology" slide before this one... Not sure if that's worth it though.

src/week1.org
```diff
@@ -0,0 +1,41 @@
...
+* The Linux terminal
+- Powerful text-based interface for interacting with your computer
+- (Almost) everything is a file
+- There's one big happy file system
+- Lots of tiny tools you can chain together to do useful things
+
+* Basic commands
+| Command | Use |
+|-------+------|
+| =cd= | Change to a different directory |
```
jgkamat: Might want to mention `pwd` as well, even though no one uses it, it's good for beginners

src/week1.org
```diff
@@ -0,0 +1,41 @@
...
+- How to contact me
+  - Slack: @kristapsberzinch
+  - Email: kristaps@robojackets.org
+
+* Why are you here?
+- Learn the basics of software development
+- Get up to speed on technologies and techniques common to most RoboJackets teams
+- TODO
+
+* The Linux terminal
```
jgkamat: Since we are not requiring people to have a VM/native installed yet, how can people test these out? Do they have to wait until teams install their VM's? We could also have them install cygwin I guess.

If most people come in with windows, most people won't be able to try out anything here.

src/week1.org
```diff
@@ -0,0 +1,41 @@
...
+|-------+------|
+| =cd= | Change to a different directory |
+| =ls=  | List files in this directory |
+| =mkdir= | Make a new directory |
+| =rm= | Remove a file |
+| =rmdir= | Remove a directory |
+| =cp= | Copy a file |
+| =mv= | Move a file |
+| =man= | Access documentation about a command |
+| =grep= | Search for a string |
```
jgkamat, line 132: a bunch of these commands are not really useful without context, for example:

`grep (pattern) [filname]` or `stdin -> grep (pattern)`. 

Maybe we could have a 'example invocation' column in this table as well, if it fits.

src/week1.org
```diff
@@ -0,0 +1,41 @@
...
+  - RoboJackets Treasurer
+  - Linux Foundation Certified Engineer
+- How to contact me
+  - Slack: @kristapsberzinch
+  - Email: kristaps@robojackets.org
+
+* Why are you here?
+- Learn the basics of software development
+- Get up to speed on technologies and techniques common to most RoboJackets teams
+- TODO
```
jgkamat: - Learn how a typical RoboJackets software project is managed
- Have a go-to place to ask questions easily
- Ensure everyone in RJ Software has a baseline software knowlege (not so sure about this one).

src/week1.org
jgkamat: Also, FWIW, I would prefer a 'tutorial walk-through' instead of a simple table, which would go through all the commands and reach a clear goal that everyone can reach. I think retention is a lot better if people feel they've applied it than just look at a table. My [git slides](https://robojackets.github.io/robocup-training/slides/2016/git) are an example of this. I don't know if we have enough time to do this though, or if it's a good idea to do this for everyone (we'll have a lot of people at this point).

src/week1.org
```diff
@@ -0,0 +1,41 @@
...
+
+* The Linux terminal
+- Powerful text-based interface for interacting with your computer
+- (Almost) everything is a file
+- There's one big happy file system
```
jgkamat: This bullet seems a little bit vague (esp if you don't have the context of a speaker), maybe say something along the lines of "All files are accessible under one 'root'".

jgkamat (dismissed):
Really great start, needs some final touches though!

src/week1.org
kberzinch: I put `pwd` on the list when we originally discussed this and it was removed in the interest of keeping this list short.

src/week1.org
kberzinch: This slide is intended as a cheat sheet for the presenter to do demos in their local shell, sort of how last year worked.

src/week1.org
kberzinch: Bash for Windows is actually pretty decent, but it takes some finagling to get up and running (about the same as cygwin imo.)

kberzinch:
Note to self: add links to these slides from readme

src/week1.org
```diff
@@ -0,0 +1,91 @@
...
+- Kristaps Berzinch
+  - RoboJackets Treasurer
+  - Linux Foundation Certified Engineer
+- How to contact me
+  - Slack: [[https://robojackets.slack.com/messages/@kristapsberzinch/][@kristapsberzinch]]
+  - Email: [[mailto:kristaps@robojackets.org][kristaps@robojackets.org]]
+
+* Who am I? - Sarah
+#+ATTR_HTML: :width 20%
+[[http://i.imgur.com/NffDOgW.jpg]]
```
jgkamat: Lookin good, but no :lock:

kberzinch:
Btw I still have several changes to make so maybe don't merge tonight (unless you want to deal with #6.)

src/week1.org
kberzinch: Per last night's meeting, we expect students to have `bash` set up in some fashion before this session, so they can follow along with the presenter.

src/week1.org
kberzinch: I'll add a note in the speaker notes for this.

src/week1.org
kberzinch: Per last night's meeting, we agreed that `rmdir` is useful.

src/week1.org
kberzinch: Per last night's meeting, the first bullet point is not something we're actually going to cover, and the last one is somewhat redundant with the first bullet point on the slide. I'll add something similar to your second bullet point though.

src/week1.org
kberzinch: Done in 6f01d65a438b04da76e07ad494086824f61416c2

src/week1.org
```diff
@@ -0,0 +1,124 @@
...
+- Jay Kamat
+  - Senior, Computer Science (Threads: Systems, Intelligence)
+  - I like contributing to Open Source!
+- Contact
+  - [[irc:irc.freenode.net:6697/jgkamat][jgkamat (freenode)]]
+  - xmpp:jayk@fysh.in
+  - [[https://matrix.to/#/@jgkamat:matrix.org][@jgkamat:matrix.org]]
+  - [[mailto:jaygkamat@gmail.com][jaygkamat@gmail.com]]
+  - [[https://github.com/jgkamat][github.com/jgkamat]]
+
```
kberzinch, line 92: @barulicm still needs to add a WAI slide.

src/week1.org
```diff
@@ -0,0 +1,124 @@
...
+#+AUTHOR: Kristaps Berzinch, Raymond Ortiz
+#+EMAIL: kristaps@robojackets.org
+
+* What are we doing today?
+- Trainer introductions
+- Overview of the RoboJackets software training program
+- Navigating and manipulating files and directories in the Linux terminal
+- Using Git and GitHub for version control
+
+* Who am I? - Kristaps
```
kberzinch: This slide gets cut off. https://puu.sh/wBsO5.png

src/week1.org
```diff
@@ -0,0 +1,124 @@
...
+- Kristaps Berzinch
+  - Junior, Computer Science (Threads: Systems Architecture, Information Internetworks)
+  - Inside RoboJackets: RoboJackets Treasurer, Former Outreach Assistant Project Manager
+  - Outside RoboJackets: 2x Capital One intern, Linux Foundation Certified Engineer
+  - a.k.a. "Mom", I occasionally give good life advice.
+- How to contact me
+  - Slack: [[https://robojackets.slack.com/messages/@kristapsberzinch/][@kristapsberzinch]]
+  - Email: [[mailto:kristaps@robojackets.org][kristaps@robojackets.org]]
+
+* Who am I? - Sarah
```
kberzinch: This slide _barely_ fits. https://puu.sh/wBsPA.png

src/week1.org
```diff
@@ -0,0 +1,124 @@
...
+- Sarah Storer
+  - Junior, Computer Science (Threads: Devices, Intelligence)
+  - Inside RoboJackets: Software Training Coordinator, Former Outreach Project Manager
+  - Outside RoboJackets: Microsoft Intern and Dog Lover
+  - a.k.a. "Dad", I dance to Hannah Montana in the shop.
+- How to contact me
+  - Slack: [[https://robojackets.slack.com/messages/@sarahstorer/][@sarahstorer]]
+  - Email: [[mailto:sstorer3@gatech.edu][sstorer3@gatech.edu]]
+
+* Who am I? - Jason
```
kberzinch: @JasonGibson274 still needs to add a pic

src/week1.org
```diff
@@ -0,0 +1,124 @@
...
+- Jason Gibson
+  - Junior, Computer Science (Threads: Devices, Intelligence)
+  - RoboJackets Training Chair
+  - IGVC Software Lead
+- How to contact me
+  - Slack: [[https://robojackets.slack.com/messages/@jasongibson/][@jasongibson]]
+  - Email: [[mailto:jgibson37@gatech.edu][jgibson37@gatech.edu]]
+
+* Who am I? - Raymond
```
kberzinch: @rortiz9 still needs to add a pic

src/week1.org
```diff
@@ -0,0 +1,124 @@
...
+- Raymond Ortiz
+  - Senior, Computer Science (Threads: Devices, Intelligence)
+  - Inside RoboJackets: IGVC Executioner, Former Training Chair
+  - Outside RoboJackets: CS1332 Head TA, Google Intern
+- How to contact me
+  - Slack: [[https://robojackets.slack.com/messages/@raymond/][@raymond]]
+  - Email: [[mailto:rortiz9@gatech.edu][rortiz9@gatech.edu]]
+
+* Who am I? - Sahit
```
kberzinch: This slide _barely_ fits. https://puu.sh/wBsQk.png

src/week1.org
```diff
@@ -0,0 +1,124 @@
...
+[[file:https://i.imgur.com/l20o7YO.jpg]]
+- Sahit Chintalapudi
+  - Sophomore, Computer Science (Threads: Intelligence, Theory)
+  - Roboracing Software Lead & Roboracing PR manager
+  - S-tier Taylor Swift fanboy
+- How to contact me
+  - Slack: [[https://robojackets.slack.com/messages/@schintalapudi/][@schintalapudi]]
+  - Email: [[mailto:schintalapudi@gatech.edu][schintalapudi@gatech.edu]]
+
+* Who am I? - Josh
```
kberzinch: This slide gets cut off. https://puu.sh/wBsQJ.png

src/week1.org
jgkamat: Unfortunately, the slides getting cut off is a longstanding problem in `reveal.js`, and I don't think they're planning on getting it fixed any time soon. AFAIK it's something about how it's hard to scale things by their height in CSS.

At any rate, the best solution I've found is to add a scrollbar to slides that overflow. 
https://stackoverflow.com/questions/35237862/best-way-to-handle-overflowing-reveal-js-slide

I can go ahead and add that, but if anyone has better CSS knowlege than me, if you come up with a diff or a css class that resizes the slides properly, I'll find a way to apply it to the slides.

That said, I think that if these slides are scaled down at all, the images would be hard to see (they're pretty hard to see right now) on a projector.

I would reccomend making the images into either vertical child slides of the about slides, or background images.  

Lastly, If anyone thinks upgrading our reveal version will help, let me know and I'll go ahead and do that. We're using 3.0.0, while 3.5.0 has been released. 

For more information, you might want to look at the [presentation size docs](https://github.com/hakimel/reveal.js#presentation-size). Right now I have the margins set to 0.22.

Reveal js targets a specific presentation size, and resizes based on that. If you overflow the presentation size, your SOL. 
2017-07-04-100647_1411x975_scrot

FWIW, if we decide to target 1080p, all the slides fit fine, since there's more than enough space, but projectors might not be 1080p, so that's proably a bad idea.

jgkamat: Oh and of course, changing the [font size](https://github.com/hakimel/reveal.js/blob/master/css/theme/black.css#L18-L20) to be smaller will also fix this (but I think it's better to have large, readable slides). Your pick though.

kberzinch: This is what that slide looks like at 1920x1080 - same problem (in Chrome, at least.)

image

I tried adjusting that margin you mentioned and it doesn't fix the issue - it just makes everything bigger but it's still the same relative amount cut off. (0.22: https://puu.sh/wATiH.png; 0: https://puu.sh/wATkL.png)

I floated the idea of background pics in Slack - I think your slide looks really good - but haven't heard back from anyone else yet.

kberzinch:
Btw here's an export of this branch rn: https://upload.kberzin.ch/uploads/week1.html

src/week1.org
jgkamat: Basically the way that reveal works is like this:

First overlay the slide contents on a specified presentation size (documented here: https://github.com/hakimel/reveal.js/#presentation-size). No resizing is done at this step, if it overflows, so be it (which is happening in this case). Next, resize the presentation window to actuallly fit on the target device. This means that as long as your slide contents are within the presentation size, you're good, but in our case, we're exceeding it, so it resizes things off the screen.

We can fix this by:
1. Increasing the margin, which I have done a bit
2. Increasing the presentation size, which gives us more room to work with (but makes everything smaller)
3. Decreasing the font size (which makes the text smaller, but keeps everything else the same).

When I was talking about the 1080p presentation size, I was talking about the 'export size' (which you can change in reveal.initialize, just for testing and playing around: https://github.com/hakimel/reveal.js/#presentation-size). 

When we do that, the slides can fit the full screen on a 1080p monitor, but when on a presentation screen, the slides get squashed (as the presentation size must be scaled down on the edges). 

Another solution that just occured to me would be to increase the presentation size without changing the aspect ratio. If you find a good presentation size, I'll go ahead and apply it to this slide deck.

1920x1080 margin 0.1
2017-07-04-103245_1920x1080_scrot

1440x1050 margin 0.1 (x1.5 resolution of what we have). 
2017-07-04-103428_1920x1080_scrot

Basically we can play around this to find a balance between readability and 'will the slides fit' for all the slides. 

Personally I think the defaults are pretty good (we should aim to limit things on a slide), but they can be annoying at times.

kberzinch: Yeah I agree. I think the actual solution here is to use background images for people pics.

kberzinch:
https://upload.kberzin.ch/uploads/week1-30cd63.html

src/week1.org
kberzinch: I ended up adjusting the presentation size and margins a bit in 2f6867dbbd6c57da49a2c95765acabd3b062b007, that seemed to work out okay.

kberzinch (commented):
@JasonGibson274 this needs to be an HTTPS link

src/week1.org
```diff
@@ -33,10 +33,13 @@
   - Email: [[mailto:sstorer3@gatech.edu][sstorer3@gatech.edu]]
 
 * Meet Jason
+#+ATTR_HTML: :width 15%
+[[http://i.imgur.com/Ty155J1.jpg]]
```
kberzinch: This needs to be an HTTPS link.

kberzinch:
https://upload.kberzin.ch/uploads/week1-44a95c1.html

ebarnette (approved):

PR #5, opened 2017-07-03
Week one example code

JasonGibson274:
I tired not to use anything that we would not expect them to know when the start training in the C++ section so it looks a little sketchy.


week1/week1.py
```diff
@@ -0,0 +1,58 @@
...
+middleInitial = 'P' #char
+yearsInRoboJackets = 3 #int
+myBirthday = 8.11 #float
+isRoboJacketsCool = True #boolean
+
+###########Lists##########
+gpas = [1.0, 2.6, 4.0, 3.6]
+membersInRoboJackets = []
+
+def exampleFunction():
```
jgkamat: Maybe this can be called `main` to keep in in line with the rest?

week1/week1.java
```diff
@@ -0,0 +1,72 @@
...
+	//This is a comment!
+	/*
+	This is a block comment...
+	By the way there are more number types like
+	byte, short, long, and float
+	but you'll run into int and double the most
+	*/
+	public int yearsInRoboJackets = 3;
+	public double myBirthday = 8.11;
+	public Boolean isRoboJacketsCool = true;
```
jgkamat: If you are trying to say that `Boolean` is passed by reference, you might want to add a comment here. Otherwise, change it to the primitive.

week1/week1.java
```diff
@@ -0,0 +1,72 @@
...
+		}
+
+		//While
+		while (length < gpas.length) {
+			System.out.println("Adjusted GPA: " + (gpas[length] + 2));
+			length++;
+		}
+
+		//Do-While
+		do {
```
jgkamat: Bumping to fix the mixed indent :(

jgkamat: Bumping on do we really need a do while?

week1/week1.py
```diff
@@ -0,0 +1,58 @@
+##########Global Variables##########
+
+###########Primitive Data Types##########
+#python isn't a typed language like C++ or Java
```
jgkamat: Python is a strongly typed language...

Python is a dynamically typed language, unlike C++ or java. (or)
Python isn't a statically typed language, like C++ or java.

week1/week1.py
```diff
@@ -0,0 +1,58 @@
+##########Global Variables##########
```
jgkamat: scripts really should have a 
`#!/usr/bin/env python` at the top.

week1/week1.py
```diff
@@ -0,0 +1,58 @@
...
+	#########Local Variables##########
+	number = 0
+	length = 0
+	count = 0
+	string = ""
+
+	##########Operators#########
+	print (3 + 3 == 9)
+	print (2 + 2 != 5)
+	print not True
```
jgkamat: Bump on adding the parenthesis :(

week1/week1.cpp
```diff
@@ -0,0 +1,77 @@
+// make sure to compile with g++
```
jgkamat: Add a makefile or a command to compile up here to make this easier.

jgkamat (changes requested):

kberzinch:
Tbh I think any Java code we put out needs to be CS 133X Checkstyle-compliant. They're going to have to get used to it anyway.

jgkamat:
cs1331 itself [doesnt](https://github.com/cs1331/cs1331.github.io/pull/1) [follow](https://github.com/cs1331/cs1331.github.io/pull/2) [checkstyle](https://github.com/cs1331/cs1331.github.io/pull/56), so I'm going to have to respectfully disagree with you on that.

kberzinch:
The code that students submit is required to be compliant, no?

week1/week1.py
JasonGibson274: should it be the actual python main method or just a method called main.

jgkamat:
Yes, but provided code has never been check-style compliant, at least in my experience (which may have changed, but still). The expectation (for me at least) is, we give them un-checkstyle compliant code, and they give us compliant code.

week1/week1.py
JasonGibson274: should it be the actual python main method or just a method called main?

jgkamat: the actual python method `__main__`? I would personally stick to just calling it `main` for readability, since it dosen't matter in our use case (I think).

JasonGibson274: I will just leave it as main

week1/week1.java
JasonGibson274: This is not for teaching just showing them the code we expect them to be able to write. I am sure some freshmen is going to used to only using do whiles. Will leave this in and talk about its at the meeting.

JasonGibson274:
@jgkamat So should I use the 1331 check style?

jgkamat:
@JasonGibson274 my vote is 👎, but you are free to do whatever you want (I'm not writing this). I feel like a style checker is useless if it does not fix your errors. Just a complaint is a waste of your time, you could be building something amazing (but instead you are busy fixing those errors).  

Please do standardize on tabs or spaces though 😄

week1/week1.cpp
```diff
@@ -0,0 +1,83 @@
...
+        num += 3;
+        cout << "Like John Cena, you can't see me" << endl;
+    } else {
+        num--;
+        cout << "Like George P. Burdell, I don't ever show up" << endl;
+    }
+
+    //////////Loops/////////
+    //For
+    for (int n = 0; n < sizeof(gpas)/sizeof(gpas[0]); n++) {
```
jgkamat, line 64: the `sizeof` logic here might be a little hard for people to grasp. Maybe we could make it easier by hard coding the length above the array (which is common in c).

week1/week1.py
```diff
@@ -0,0 +1,66 @@
...
+    print (3 + 3 == 9)
+    print (2 + 2 != 5)
+    print (not True)
+    print (True or False)
+    print (True and False)
+
+    #########Loops#########
+    ##If-ElseIf-Else
+    if (number == 0):
+    number = 3
```
jgkamat: Indentation :(

week1/week1.py
```diff
@@ -0,0 +1,66 @@
...
+    print ("I understand Loops!")
+    elif (number > 4):
+    number += 3
```
jgkamat: Indentation

week1/week1.py
```diff
@@ -0,0 +1,66 @@
...
+    print ("Like John Cena, you can't see me")
+    else:
+    number -= 1
```
jgkamat: Indentation

week1/week1.py
```diff
@@ -0,0 +1,66 @@
...
+    print ("Like George P. Burdell, I don't ever show up")
+
+    ##For
+    for n in gpas:
+    print ("GPA:", n)
```
jgkamat: Indentation

week1/week1.py
```diff
@@ -0,0 +1,66 @@
...
+
+    ##While
+    length = 0
+    while (length < len(gpas)):
+    print ("Adjusted GPA:", (gpas[length] + 2))
```
jgkamat: Indentation

week1/week1.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
+
+# To run this file do
+#		python3 week1.py
+
+##########Global Variables##########
```
jgkamat: There's nothing under this header, so it's a bit confusing. I would remove it, or explain how everything in the root is global (to the module).

Global is kind of weird in python since you can access everything if you want...

week1/week1.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    //Do-While
+    do {
+        cout << "Count is: " << count << endl;
+        count++;
+    } while (count < 11);
+
+    // this is a status code that will be returned to whatever ran this executable
+    // 0 typically means to errors
```
jgkamat: `0` means success in exit codes. `0` means false in if statements though... :cry:

week1/week1.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        //While
+        while (length < gpas.length) {
+            System.out.println("Adjusted GPA: " + (gpas[length] + 2));
+            length++;
+        }
+
+        //Do-While
+        do {
```
jgkamat, line 70: Bumping do-while again.

We have 0 instances of do-while across all RJ codebases that I know about currently, so I don't think this is needed at all. Let me know if you think differently though.

jgkamat (dismissed):
Your python is broken, but other than that, a bunch of nitpicks

week1/week1.cpp
ebarnette: I'm fine with this logic....it motivates them to use the std lib way of doing it which they will learn later

week1/week1.java
ebarnette: We made an executive decision to leave them in

jgkamat (approved):
Could you add `*.class` to the .gitignore?

That's the only issue I have, so I'm :+1:'ing this preemptively.

jgkamat:
Also one more suggestion, maybe we can move the code to a seperate subfolder, so every week dosen't clutter up the root? 

Maybe `code/week1`. I could also rename the `src` folder to something else if you want to take that.

PR #6, opened 2017-07-04
Restructures repo. May break things.

ebarnette:
@jgkamat I probably broke a lot of slide things.

also TEST.org is in the code file as a placeholder


ebarnette:
Also CI is going to be very broken.

jgkamat:
@sastorer I feel like this is a bad idea right now, because I think this might cause a bunch of annoying conflicts in #4 and #5. Can we merge them as is for now, and I'll submit a PR after reorging the repo?

ebarnette:
Yeah sure.

I want:
a) as few top level directories as possible (ideally just Slides and Code)
b) folders under both Slides and Code directories labeled by week

jgkamat:
@sastorer sounds good, I'll plan for a 
- `./slides/1`
- `./code/1`

I'd rather stick to lowercase, since it follows the convention a bit more, but if you *really* want uppercase, I'll do it.

Also, just FYI, most of the files you moved from the root probably should to stay in the root (eg: the makefile, cask, etc), to make it super easy for newcomers to build the slides. The source files themselves (*.org, *.{c,java,py}) we can move wherever.

kberzinch:
I vote lowercase but slide-related files in a slides folder.

ebarnette:
@jgkamat just merged example code

Waiting on the WAI section of topic slides to merge and then you can do your reorg magic...it'll just look gross for like another day

ebarnette:
@jgkamat You're good to do this ASAP. Just merged the topic slides from week 1.

PR #8, opened 2017-07-05
Undeleted raymond's pic

ebarnette:


jgkamat:
This should be merged right? :3

Nevemind :P

ebarnette:
I fixed these problems by using admin powers:P

I accidentally deleted his pic on my imgur account so the link went nowhere...

PR #14, opened 2017-07-06
Fix md image export

jgkamat:
Not too big of a chnage, shouldn't break anything AFAIK. 

Let me know if you spot anything fishy.


kberzinch (dismissed):

jgkamat:
I actually decided to overload this with a small change to remove old artifacts, right now old artifacts are just hanging around. Probably wont break anything :crossed_fingers: 

EDIT: actually nvmd, give me a bit.
EDIT2: should be gtg now.

kberzinch (approved):

jgkamat:
That actually didn't remove the old artifacts, but they aren't hurting anything, so I'll leave them as is for now, and fix them in my next PR.

PR #15, opened 2017-07-09
Week two git slides

joshhting:


slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
+#+TITLE: GitHub Training Two 2017
+#+AUTHOR: Sahit Chintalapudi, Josh Ting 
+#+EMAIL: schintalapudi@gatech.edu, josh.ting@gatech.edu
+
+* Let's Recap
+- We have a fork of the software training repo on our local GitHub account
+- Reminder: a fork is an independently maintained repository based off of another
```
jgkamat: Maybe you can just say 'a fork is a (deep) copy of another git repository'

'based off of another' feels weird and can be confusing

slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
...
+- The original repository is referred to as "upstream"
+
+* Things have changed upstream, what do we do?
+- We need a series of git commands that can update our fork and local repo with the new content in the upstream repo
+- All terminal git commands are in the form:
+  #+BEGIN_SRC shell
+    git <command> <param1> <param2> ...
```
jgkamat, line 15: do you want me to make `git` a special keyword so it turns green? That would be pretty cool :P

slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
...
+  #+END_SRC
+
+* Git pull
+- First, navigate to the git repo from the terminal
+- We can keep our fork up to date with the /git pull/ command
```
jgkamat: `/` should probably be `~`

slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
...
+        your project not behaving properly
+- Anything  commited to git stays in git. 
+      - This doesn't mean bad commits can't be reverted
+      - This does mean that you should never commit things like passwords to
+        git
+
+# * Good vs bad:
+# | Good commit message | Bad commit message |
+# |---------------------+--------------------|
+# | [[file:https://i.imgur.com/k0Llu0s.png]] | [[file:https://i.imgur.com/8vCOZQP.png]] |
```
jgkamat: :100:

slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
...
+    cd software-training    # or wherever your repo is
+    git pull upstream master
+  #+END_SRC
+
+- What does this mean?
+      - /git pull/: pull from one GitHub repository into our current repository
+      - /upstream/: specify that we are pulling from the upstream repository
+      - /master/: pull into the "master" branch of our current repository
+- What are branches? That's classified! Stay tuned for next lesson
+# Basically say that this is outside the scope of the current lesson and if they want to read up on it on their own we won't stop them
```
jgkamat: I think that if the remote branch is the same as the local, you can specify `git pull upstream`. That depends on your gitconfig though (`push.default`), so I would lean towards keeping it this way. Just letting you know that could be an option if we want to ignore branches completly right now.

slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
...
+
+* Let's make some of our own changes!
+- Open up a text editor of your choice and create a new file with some text
+  in it
+- Save the file inside the repository
+- Now how do we make this file show up online for others to see?
+
+* Time to make them official:
+- Git encapsulates changes in "commits"
```
jgkamat: This sounds a little confusing, maybe:

Git calls every version it stores a 'commit'

slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
...
+- /master/: Name of the branch to push to(Still top secret material)
+
+* Some notes about commits 
+[[file:https://imgs.xkcd.com/comics/git_commit.png]]
+- A good commit message is short but clearly explains what changes were made
+      - A good commit message makes it easy to see what changes could lead to
+        your project not behaving properly
+- Anything  commited to git stays in git. 
+      - This doesn't mean bad commits can't be reverted
+      - This does mean that you should never commit things like passwords to
```
jgkamat: Also add a bullet about large files :)

slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
...
+- /git push/: Command to push commits to another repository
+- /origin/: Name of the repo to push to (origin is referring to our fork)
+- /master/: Name of the branch to push to(Still top secret material)
+
+* Some notes about commits 
+[[file:https://imgs.xkcd.com/comics/git_commit.png]]
+- A good commit message is short but clearly explains what changes were made
+      - A good commit message makes it easy to see what changes could lead to
+        your project not behaving properly
+- Anything  commited to git stays in git.
```
jgkamat: kind of a lie, but I'd excuse it for newcomers. If there's a better way of wording this that doesn't imply you can never change history I think we should go with that.

Case study: Someone creates a branch, puts important work on it, and deletes it. Then they wait 30d and it gets garbage collected, and they want their work back. According to this slide, they should be able to, but in reality, they can't.

(in mercurial, there is no garbage collection, so this statement would be true).

slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
...
+
+* Time to make them official:
+- Git encapsulates changes in "commits"
+- In order to commit something, you must specify which changes you've made to the codebase should be in the commit
+- This is called "staging" your changes
+
+* Git add
+
+  #+BEGIN_SRC shell
+    git add .
```
jgkamat, line 48: No one will see the `.` at first. Maybe add a line comment above it with something like

`# git add <PERIOD>`

I thnk `git add -A` would work too, but not sure.

slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
...
+
+* Git commit
+
+  #+BEGIN_SRC shell
+    git commit -m "Added a file!"
+  #+END_SRC
+
+- /commit/: Commit currently staged changes to git
+  - This is making the change "permanent" (more on this later)
+- /-m "..."/: Commits require commit messages to label them
```
jgkamat: We should also make sure we tell them `-<letter>` and `--<word>` are ways for specifying options for cli programs. I'm not sure if we covered this earlier though.

jgkamat (changes requested):
Overall looking great! Nice job! :tada:

kberzinch:
This is supposed to merge into `weekTwo` not `master`.

slides/week2/git.org
joshhting: That'd be neat

jgkamat:
You probably also want to update the README to add these slides.

kberzinch:
I have a tangentially related nitpick - week 1 git slides were `github.org`, this week they're `git.org`. I propose we standardize on one or the other.

slides/week2/git.org
kberzinch: Seconded

slides/week2/git.org
kberzinch: also bumping same tweak from earlier - use `~` instead of `/` for command snippets

slides/week2/git.org
```diff
@@ -0,0 +1,88 @@
...
+
+* Git push
+
+  #+BEGIN_SRC shell
+    git push origin master
+  #+END_SRC
+
+- /git push/: Command to push commits to another repository
+- /origin/: Name of the repo to push to (origin is referring to our fork)
+- /master/: Name of the branch to push to(Still top secret material)
```
kberzinch: Missing a space before the last open parenthesis

kberzinch (changes requested):
Typo

joshhting:
I'll rename the week 1 file to git.org then. And fix the title slides to reflect that

slides/week2/git.org
```diff
@@ -0,0 +1,101 @@
...
+- We can keep our fork up to date with the ~git pull~ command
+
+  #+BEGIN_SRC shell
+    cd software-training    # or wherever your repo is
+    git pull upstream master
+  #+END_SRC
+
+- What does this mean?
+      - ~git pull~: pull from one GitHub repository into our current repository
+      - ~upstream~: specify that we are pulling from the upstream repository
```
JasonGibson274: Just a note. Last week they named their upstream rj.

slides/week2/git.org
```diff
@@ -0,0 +1,101 @@
...
+    your project not behaving properly
+
+* Some notes about commits cont.
+- Things committed to Git are intended to stay as a permanent record of the repository history 
+  - This doesn't mean bad commits can't be reverted
+  - This does mean that you should never commit things like passwords to
+    git
+  - This does mean that good commit messages are important
+- Large files are impractical for Git to track
+  - Git will refuse to push files larger than 100 MB.
```
jgkamat: Er, this isn't actually part of git. This is a limitation of github. Git will commit/push such files just fine (but might be slow). For example, [gitlab has a 10gb overall limit](https://about.gitlab.com/2015/04/08/gitlab-dot-com-storage-limit-raised-to-10gb-per-repo/) (and no hard per-file limit).

jgkamat (dismissed):
Have a few nitpicks, but everything else looks good so conditional :+1:

kberzinch (approved):

jgkamat (approved):

PR #16, opened 2017-07-10
Week two example code

ebarnette:
Adds week 2 example code to weekTwo branch


code/week2/operations.h
```diff
@@ -0,0 +1,9 @@
+#ifndef OPERATIONS_H
```
jgkamat: are we not going to tell them about `#pragma once`? This is pretty much the biggest reason to use cpp over c.

code/week2/operations.cpp
```diff
@@ -0,0 +1,25 @@
...
+    return a + b;
+}
+
+int subtract(int a, int b) {
+    return add(a, -b);
+}
+
+int multiply(int a, int b) {
+    int accum = 0;
+    for(int i = 0; i < a; i++) {
```
jgkamat: Is there a particular reason you're not just using `*`? This will probably be slower on most compilers, and is more complex to read.

code/week2/week2.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    // maximum length a size can reach due to system limitation. DO NOT try to reach this size.
+    cout << str1.max_size() << endl;
+
+    str1.resize(3);
+    cout << str1 << endl;
+    str2.resize(10, 'a');
+    cout << str2 << endl;
+
+    string str3 = "";
```
jgkamat: The comments stopped at this point 😢 

At least put links to the cpp docs

code/week2/week2.cpp
```diff
@@ -0,0 +1,85 @@
+// make sure to compile with g++
+// to compile do
+//      g++ -o week2.out week2.cpp operations.cpp
```
jgkamat: We can probably make this 
`g++ -o week2.out *.cpp`
so that the command is shorter and we don't need to change it all the time.

code/week2/week2.cpp
```diff
@@ -0,0 +1,85 @@
...
+    // I am actually using do_something function from library B) and imrpoving readaility of the code.
+    // decent stack overflow article https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice
+    using namespace std;
+
+    // ***************** FUNCTIONS **************
+
+    cout << "2 + 2 = " << add(2, 2) << endl;
+    cout << "2 - 4 = " << subtract(2, 4) << endl;
+    cout << "2 * 2 = " << multiply(2, 2) << endl;
+    cout << "6 % 2 = " << divide(6, 2) << endl;
```
jgkamat: `%` is not `/`

jgkamat (changes requested):
Overall, looking good! Nice job! 😄

code/week2/operations.cpp
JasonGibson274: This is to show chaining of functions

JasonGibson274: I can change it if that is dumb

ebarnette: I like the chaining of functions in the example (also example code shouldn't reeeeeeally always be optimized for the compiler)

jgkamat: While 'we shouldn't be optimizing' is normally the case for example slides, this is a bit extreme. This 5 line cpp function can be replaced by a _single_ line of x86. The easiest assembly implementation of this function is simpler, easier to understand, and orders of magnitude more efficient than this cpp function.

This sets a dangerous precedent because:

1) this implies cpp does not have native multiplication
2) It encourages roundabout, convoluted solutions when simple, obvious solutions are present
3) (if someone already knows cpp): it makes RJ look like amateurs 
4) It's a waste of student time to parse and understand this function (ie: what's going on here, is this just multiply? Am I missing something? There has to be a reason it's implemented with adds... What am I missing?).

I would suggest actually making a function that has a purpose, rather than making a  purely trivial example. Maybe implement a exponent function instead, using a wrapper around multiply? While `std::pow` does exist, at least this can be excused as an 'alternate implementation' of `std::pow`, rather than a re-implementation of a cpp primitive. Even better, maybe implement a looping factorial (which, afaik, isn't in `std`).

JasonGibson274: This was originally done just to show that you can call a function from inside a function and how chaining them together can produce useful code. Remember that this may be the first time these students are introduced to functions at all and they are expected to following along in the example code as the lectures progresses. We can change the four functions to a one lined multiply with printing the numbers so it is less trivial and an iterative implementation of factorial.

JasonGibson274: Then the print messages will allow them to track function calls as they happen.

ebarnette: @jgkamat We discussed this in Monday's meeting and decided that it would be fine

ebarnette:
Needs examples for operator and function overloads. Also pending @kberzinch for further review.

code/week2/operations.cpp
kberzinch: You could also add a comment on this function to that effect.

JasonGibson274:
Still missing overloading operators and converting primitive to strings and string to primitives. I started using \n for formatting but If we want to standardize to endl instead just comment. I left in the multiply function as of right now. Hopefully the print statements make the intention more clear.

JasonGibson274:
@kberzinch The example has everything covered in the topic slides. Just needs a final look through and it should be done.

code/week2/week2.cpp
```diff
@@ -0,0 +1,125 @@
...
+//      ./week2.out
+
+// This line includes the week2 header file defined in this directory
+#include "week2.h"
+
+int main() {
+    // This means that this code uses the std namespace, meaning all of the functions in the std namespace
+    // do not require a qualifier. Generally using namespaces is considered bad practice especially in large
+    // projects since it can lead to shadowing (where I think I using do_something fucntion from library A but
+    // I am actually using do_something function from library B) and imrpoving readaility of the code.
```
kberzinch: This sentence needs help

code/week2/week2.cpp
```diff
@@ -0,0 +1,125 @@
...
+    // decent stack overflow article https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice
```
kberzinch: Shorter SO link: https://stackoverflow.com/q/1452721

code/week2/week2.cpp
```diff
@@ -0,0 +1,125 @@
...
+    // empty checks if the string is empty ""
+    string str3 = "";
+    if(str3.empty()) {
+        cout << "str3 is empty" << endl;
+    } else {
+        cout << "str3 is NOT empty" << endl;
+    }
+
+    // http://www.cplusplus.com/reference/string/string/clear/
+    // clear repalces the string with the null string ""
```
kberzinch: Minor typo

kberzinch (changes requested):
I also dislike the inconsistency in comment layout. Please either put the link first or the description first consistently throughout week2.cpp.

ebarnette:
@kberzinch and @jgkamat your requested changes have been addressed

code/week2/week2.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+// This line includes the week2 header file defined in this directory
+#include "week2.h"
+
+int main() {
+    // This means that this code uses the std namespace. Now, all of the functions in the std namespace
+    // do not require a qualifier. Generally using namespaces is considered bad practice especially in large
+    // projects since it can lead to shadowing (where I think I am using do_something fucntion from library A but
+    // I am actually using do_something function from library B). Cutting down on extra declarations
+    // imrpoves the readaility of the code.
```
kberzinch: Still two typos on this line.

kberzinch (changes requested):

ebarnette:
@kberzinch happy now?

kberzinch (dismissed):

code/week2/operations.cpp
```diff
@@ -0,0 +1,51 @@
...
+    }
+    return accum;
+}
+
+// divide
+int divide(int a, int b) {
+    std::cout << "dividing (int) " << a << " / (int) " << b << std::endl;
+    int result = 0;
+    while(a > 0) {
+      a = subtract(a, b);
```
jgkamat: This blows up if `b` is negative.

code/week2/operations.cpp
```diff
@@ -0,0 +1,51 @@
...
+      accum = add(accum, b);
+    }
+    return accum;
+}
+
+// divide
+int divide(int a, int b) {
+    std::cout << "dividing (int) " << a << " / (int) " << b << std::endl;
+    int result = 0;
+    while(a > 0) {
```
jgkamat: This causes an incorrect answer when `a` is < 0.

code/week2/operations.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+#include <iostream>
+
+int add(int a, int b);
```
jgkamat: Header files usually have method documentation, maybe this file should have docs?

code/week2/operations.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+// add
+int add(int a, int b) {
+    std::cout << "adding (int) " << a << " + (int) " << b << std::endl;
+    return a + b;
+}
+
+int add(int a, int b, int c) {
+    std::cout << "adding (int) " << a << " + (int) " << b << " + (int) " << c << std::endl;
+    return a + b + c;
```
jgkamat: This should be `add(a, add(b, c))` for consistency.

jgkamat (changes requested):

JasonGibson274:
@jgkamat should be done now. Sorry I thought this pull request was already merged.

code/week2/operations.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+    if (neg) {
+        return -accum;
+    } else {
+        return accum;
+    }
+}
+
+// divide
+int divide(int a, int b) {
```
jgkamat, line 52: Divide infinite loops when dividing by 0, but that's ok, since c standard says division by 0 is undefined.

jgkamat (approved):

code/week2/operations.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+    if(b < 0) {
+        b = -b;
+    }
+    while(a > 0) {
+      a = subtract(a, b);
+      result++;
+    }
+    if(neg) {
+      return -result;
```
jgkamat: Actually, you have mixed indent levels in your file.

Choose between 4 and 2 spaces please :smile:

jgkamat (changes requested):

code/week2/operations.cpp
JasonGibson274: added comment to .h file

jgkamat (dismissed):

code/week2/week2.cpp
```diff
@@ -0,0 +1,126 @@
...
+// to run do
+//      ./week2.out
+
+// This line includes the week2 header file defined in this directory
+#include "week2.h"
+
+int main() {
+    // This means that this code uses the std namespace. Now, all of the functions in the std namespace
+    // do not require a qualifier. Generally using namespaces is considered bad practice especially in large
+    // projects since it can lead to shadowing (where I think I am using do_something fucntion from library A but
```
kberzinch: Typo: `fucntion`

kberzinch (changes requested):
One last typo

jgkamat (approved):

kberzinch (approved):

PR #18, opened 2017-07-11
Minor week1 fix

joshhting:


jgkamat (approved):
LGTM!

This README changing nonsense is pretty annoying, maybe I can try to generate a higher level TOC so we don't need to keep updating it, the same way my website currently links pages. I'll see if I can get some time to do that though..

kberzinch (approved):

PR #27, opened 2017-07-14
Vendor Reveal.js on our own domain

### LINKED ISSUE #20: Speaker notes aren't working
Pressing `S` on a slideshow opens a popup but the HTML source is displayed rather than the page content.

Tested on latest Chrome and Edge on latest Windows.

jgkamat:
This moves reveal.js from the CDN we were using previously onto rj.github, which solves some issues with speaker notes. 

I also attempted to clean up old html files in the auto-exporter, so we have to be careful when merging this.

Closes #20

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,5 +24,5 @@ jobs:
           name: Deploy to github
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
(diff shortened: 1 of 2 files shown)


kberzinch (approved):

jgkamat:
Going to hold off on merging this because it has the potential to break some things and I'd rather be at home for dealing with that. Scheduling for a merge tonight.

PR #30, opened 2017-07-19
Week three topic slides

kberzinch:
This still needs some work, but I'll go ahead and put this out there. Feedback appreciated as always.

In particular, I'd like advice on any speaker notes to add to the slides.

### DIFF
--- a/slides/week3/week3.org
+++ b/slides/week3/week3.org
@@ -2,6 +2,8 @@
 #+AUTHOR: Kristaps Berzinch, Raymond Ortiz
 #+EMAIL: kristaps@robojackets.org, rortiz9@gatech.edu
 
+* TODO Needs speaker notes
+
 * What are we doing today?
 - Data structures and algorithms in C++
 - Brancing and merging in Git
--- a/slides/week3/week3.org
+++ b/slides/week3/week3.org
@@ -39,7 +39,9 @@
 | =pop_back= | removes the last element |
 
 * Iterators
--
+- An object that lets you access objects in a container
+- Depending on their type, they will let you read or write data in the container, and move forwards, backwards, or arbitrarily
+- The containers library includes the most common ones you'll need, but you can also write your own
 
 * Algorithms
 - All of these can be restricted to a range in the container
--- a/slides/week3/week3.org
+++ b/slides/week3/week3.org
@@ -6,7 +6,7 @@
 
 * What are we doing today?
 - Data structures and algorithms in C++
-- Brancing and merging in Git
+- Branching and merging in Git
 
 * What's a constructor?
 - A special type of function that instantiates a class
@@ -14,16 +14,23 @@
 
 * What's a container?
 - An object you can use to store other objects
-- They can be ordered, associative, both, or neither
-  - *Ordered:* The contents are in a specific order
-  - *Associative:* Two objects are associated with each other within the container
-- TODO am I missing any important properties here? I want to say "uniqueness" or something...
+- Objects are identified by keys (which can also be objects)
+- Some important properties:
+  - *Ordered:* the contents are in a specific order
+  - *Associative:* two objects are associated with each other within the container
+  - *Set:* the value of an object is its key
+  - *Unique keys:* no two objects can have the same key
 
 * Common containers
+#+BEGIN_NOTES
+Note that in most languages, sets and maps are unorded. C++ also has unordered sets and maps.
+#+END_NOTES
 | =array= | a static-length ordered list of objects |
 | =vector= | a dynamic-length ordered list of objects |
 | =set= | an ordered group of unique objects |
-| =map= | an ordered group of key-value pairs, with unique keys|
+| =unordered_set= | an unordered group of unique objects |
+| =map= | an ordered group of key-value pairs, with unique keys |
+| =unordered_map= | an unordered group of key-value pairs, with unique keys |
 
 * Arrays
 | =at= | returns the value at a specific index |
@@ -32,6 +39,9 @@
 | =size= | returns the number of elements |
 
 * Vectors
+#+BEGIN_NOTES
+Vectors can grow as items are added, whereas arrays are static.
+#+END_NOTES
 - Everything from the last slide, and...
 | =clear= | removes all items |
 | =insert= | inserts an item at a specific index |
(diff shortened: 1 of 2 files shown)


slides/week3/week3.org
```diff
@@ -0,0 +1,54 @@
+#+TITLE: Week 3
+#+AUTHOR: Kristaps Berzinch, Raymond Ortiz
+#+EMAIL: kristaps@robojackets.org, rortiz9@gatech.edu
+
+* TODO Needs speaker notes
+
+* What are we doing today?
+- Data structures and algorithms in C++
+- Brancing and merging in Git
```
jgkamat: Branching

slides/week3/week3.org
```diff
@@ -0,0 +1,54 @@
...
+* What's a constructor?
+- A special type of function that instantiates a class
+- Basically, it sets up an object and returns it
+
+* What's a container?
+- An object you can use to store other objects
+- They can be ordered, associative, both, or neither
+  - *Ordered:* The contents are in a specific order
+  - *Associative:* Two objects are associated with each other within the container
+- TODO am I missing any important properties here? I want to say "uniqueness" or something...
```
jgkamat: From c++.com: sets


>Associative
    Elements in associative containers are referenced by their key and not by their absolute position in the container.
Ordered
    The elements in the container follow a strict order at all times. All inserted elements are given a position in this order.
Set
    The value of an element is also the key used to identify it.
Unique keys
    No two elements in the container can have equivalent keys.
Allocator-aware
    The container uses an allocator object to dynamically handle its storage needs.

slides/week3/week3.org
```diff
@@ -0,0 +1,54 @@
...
+
+* Common containers
+| =array= | a static-length ordered list of objects |
+| =vector= | a dynamic-length ordered list of objects |
+| =set= | an ordered group of unique objects |
```
jgkamat: We should at least make a verbal note that sets in other languages are unordered. Maybe we could teach them about unordered_set instead?

slides/week3/week3.org
```diff
@@ -0,0 +1,54 @@
...
+| =map= | an ordered group of key-value pairs, with unique keys|
```
jgkamat: same comment as above, AFAIK, unordered map is used a lot more and has better performance.

slides/week3/week3.org
```diff
@@ -0,0 +1,54 @@
...
+- The containers library includes the most common ones you'll need, but you can also write your own
+
+* Algorithms
+- All of these can be restricted to a range in the container
+| =sort= | sorts a container |
+| =count= | counts the number of items in a container that match a given item |
+| =find= | returns an iterator to the first element that matches a given item |
+| =copy= | copies elements from one container to another |
+| =fill= | fills a container with copies of a given element |
+| =accumulate= | sums all elements in a container |
```
jgkamat, line 65: Transform (map) is very useful when paired with accumulate (reduce) (eg: transform to multiply all elements by 2, then accumulate to sum/subtract/multiply them together).

slides/week3/week3.org
kberzinch: I don't think we need or want to talk about allocation this week.

I would like to frame "set" and "unique keys" as properties of a collection but I'm not sure how to phrase that.

slides/week3/week3.org
kberzinch: I'll add to speaker notes.

RE unordered_containers: we can mention that each container (I think?) has an unordered version. We should teach unordered_set just to show how that works with iterators.

slides/week3/week3.org
JasonGibson274: Do we want to be more explicit which headers these come from?

slides/week3/week3.org
```diff
@@ -0,0 +1,54 @@
...
+- Basically, it sets up an object and returns it
+
+* What's a container?
+- An object you can use to store other objects
+- They can be ordered, associative, both, or neither
+  - *Ordered:* The contents are in a specific order
+  - *Associative:* Two objects are associated with each other within the container
+- TODO am I missing any important properties here? I want to say "uniqueness" or something...
+
+* Common containers
```
JasonGibson274: Are we planning on covering map and set? Also we would need to cover either iterators or for each loops before we cover these two since there is not a good way to print them out without using either.

slides/week3/week3.org
jgkamat: Here's an attempt 😢 
Containers hold pieces of data
Index: You need a way to index into any container, be it an assigned number, the data itself, or another value you provide. This is called the 'Key' or 'Index' of a piece of data.
Unique Keys: No two data entries can have the same index. If they could, how would C++ decide which piece of data to give you?

slides/week3/week3.org
```diff
@@ -0,0 +1,54 @@
...
+| =set= | an ordered group of unique objects |
+| =map= | an ordered group of key-value pairs, with unique keys|
+
+* Arrays
+| =at= | returns the value at a specific index |
+| =front= | returns the first item |
+| =back= | returns the last item |
+| =size= | returns the number of elements |
+
+* Vectors
```
JasonGibson274, line 41: We should make a note about how vectors can grow in size with arrays are static. I am assuming this will be in the speaker notes.

JasonGibson274 (changes requested):

slides/week3/week3.org
JasonGibson274: set and map have unordered versions. I do not think vector or arrays do.

slides/week3/week3.org
```diff
@@ -0,0 +1,54 @@
...
+- Everything from the last slide, and...
+| =clear= | removes all items |
+| =insert= | inserts an item at a specific index |
```
JasonGibson274: To my knowledge Index is given by an iterator.

slides/week3/week3.org
kberzinch: @JasonGibson274 I tried that just now - adding a column sort of wrecks the slide layout. I'll add a note in the speaker notes though.

slides/week3/week3.org
kberzinch: I'm not sure if we want to discuss them in much detail, but students should at least be aware they exist at this stage. I'm comfortable leaving them in this table but not discussing them any further.

We're using iterators this week anyway.

slides/week3/week3.org
kberzinch: Yes, but we're not quite at iterators in this deck.

slides/week3/week3.org
kberzinch: Added.

slides/week3/week3.org
kberzinch: @jgkamat added.

JasonGibson274 (approved):

PR #34, opened 2017-07-25
Week four topic slides

kberzinch:


slides/week4/week4.org
```diff
@@ -0,0 +1,87 @@
...
+
+* Type Safety
+#+BEGIN_SRC shell
+      char c = 10;                        // this is one byte in memory
+
+      int *p = (int*) &c;                 // this is a 4-byte pointer pointing to one byte of
+                                          // memory - it compiles but leads to corrupted memory
+                                          // if you try to write to what p points to
+
+      int *q = static_cast<int*> &c;      // throws an exception at compile time
```
JasonGibson274: missing ()
int *q = static_cast<int*>(&c);

JasonGibson274 (dismissed):

slides/week4/week4.org
```diff
@@ -0,0 +1,87 @@
...
+- You should generally use one of these two types instead of a normal pointer:
+ - =shared_ptr=: multiple =shared_ptr= s can own a single object, and the object is destroyed when the last =shared_ptr= is destroyed.
+ - =unique_ptr=: only one =unique_ptr= can own an object, and the object is destroyed when the =unique_ptr= is destroyed.
+- You should initizialize these using the =make_shared= and =make_unique= functions, respectively
+
+* Casting, the wrong way
+- Sometimes you want to convert a variable to a different type: this is called casting
+- In C, you can simply change the type of a variable like so:
+ - ~int a = (int)not_an_int;~
+ - Now =a= is an =int= reference to =not_an_int=, which may or may not actually be an =int=
```
barulicm: _a_ is not a reference in this example. 
A better example for this might involve pointers since we've taught them that.
B *old_var = ...;
A *new_var = (A*)old_var;
Here, the compiler happily learns to treat whatever object is at the address in old_var as an object of type A. It will use the same bytes of data, but interpret them differently. This can be semantically meaningless depending on the specific types A and B.

For a more detailed example, which is too large for a slide, consider the following structs
struct A {
short member1;
};
struct B {
char member1;
char member2;
};
A *a = ...;
B *b = (B*)a;

A novice programmer may mistakenly think that the compiler will somehow know how to transform between these two types, not realizing that the compiler will implicitly just split the 2 bytes of a->member1 into b->member1 and b->member2. This will likely not have any semantic meaning in the domain of the application.
While this is a trivial example, as our classes become more complex, even more experienced programmers can mistakenly cast between incompatible types. the *_cast<>() functions discussed this week offer checks to ensure that various bad things like this don't happen.

barulicm (commented):
I think these explanations are handled well, barring the one line I commented on. 
These slides feel very word heavy, but I haven't looked at any of the other weeks' topic slides yet.

kberzinch:
Yeah, this week's slides are wordier than usual. I'll look into moving more content into speaker notes.

barulicm (approved):

jgkamat:
Any reason this was merged into master? :confused:

kberzinch:
Fixed in #40, #41

jgkamat:
er, I meant that this branch is currently in master. I don't mind if that's not an issue though:

https://github.com/RoboJackets/software-training/blob/master/slides/week4/week4.org

PR #35, opened 2017-07-26
Week four example code

JasonGibson274:
I just wanted to open this up for review early. I was planning on adding more to the pointer section to get more depth in examples.

### DIFF
--- a/code/week4/week4.cpp
+++ b/code/week4/week4.cpp
@@ -1,3 +1,8 @@
+/*
+ * to compile
+ * g++ -std=c++14 -o week4.out week4.cpp
+*/
+
 #include "week4.h"
 
 void no_ptr_method(int num) {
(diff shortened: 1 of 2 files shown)


JasonGibson274:
I will remove some of the explanation depending on the topic slides but if you want any removed just comment. I know this throws a compiler warning with returning a pointer to a local variable but the first time I learned what a pointer was I made this exact mistake so I think it is worth covering explicitly. I can also remove the vector of pointers if looks to confusing.

code/week4/week4.cpp
```diff
@@ -0,0 +1,199 @@
...
+    /**********************************************************************
+    *   OPERATOR  |               MEANING                                 *
+    * ------------------------------------------------------------------- *
+    *     VAR.    | accesses a class member given the actual object       *
+    *     VAR->   | accesses a class member given a pointer to the object *
+    ***********************************************************************/
+	// to reference the underlying vector you should use the -> operator
+	// it is just a short hand for (*()).
+	vector_a_ptr->insert(vector_a_ptr->begin(), "too");
+	// see why -> exists, Thanks Dennis Ritchie for having an eye for sytle
```
kberzinch: style

kberzinch (changes requested):
Other than that, looks good to me!

kberzinch (approved):

PR #36, opened 2017-07-29
Week five topic slides

kberzinch:
The Trello card on this was really brief - I fleshed out what I could think of but this still seems really short.

Advice on expanding this deck somewhat is appreciated.

### DIFF
--- a/slides/week5/week5.org
+++ b/slides/week5/week5.org
@@ -75,8 +75,8 @@ class child: public example {
  #+END_SRC
  #+BEGIN_SRC
  example.cpp: In function ‘int main()’:
- example.cpp:20:38: error: cannot dynamic_cast ‘e’ (of type ‘class example’) to type ‘class child&’
- (source type is not polymorphic)
+ example.cpp:20:38: error: cannot dynamic_cast ‘e’ (of type ‘class example’) to type
+ ‘class child&’ (source type is not polymorphic)
       child& c = dynamic_cast<child&>(e);
                                       ^
  #+END_SRC
--- a/slides/week5/week5.org
+++ b/slides/week5/week5.org
@@ -4,7 +4,7 @@
 
 * What are we doing today?
 - Classes in C++
-- Merge conflicts in =git=
+- Handling merge conflicts in =git=
 
 * What are classes?
 - Definitions for objects, i.e. what data and methods they have
@@ -16,12 +16,17 @@ A class can have some data and some methods, which can either be public or priva
 #+END_NOTES
 #+BEGIN_SRC c
 class example {
-    private:                        // private is also implied if you don't specify visibility
+    private:         // private is also implied if you don't specify visibility
         int a;
     public:
-        int getA() { return a; }    // why wouldn't we just make `a` public?
+        int getA() { return a; }
 };
 #+END_SRC
+
+* Why bother?
+- Software projects are easier to understand when related data and functions are grouped together
+- Abstracting data behind getters/setters allows you to validate inputs from other parts of your application
+- Abstracting complex tasks into class methods enables simpler, easier to read high-level code
 
 * Static members
 #+BEGIN_NOTES
@@ -30,12 +35,45 @@ class example {
 - For when you want data or functions to be part of your class, but they don't need to be "attached" to an instance of the class
 - Remember the =static= keyword means something else outside of class definitions!
 #+BEGIN_SRC c
-class example {
+class static_example {
     private:
         static int a;
     public:
         static int getA() { return a; }
 };
+#+END_SRC
+
+* Inheritance
+- Classes can inherit members from other classes
+#+BEGIN_SRC c
+class child: public example {
+    // we get `a` and `getA()` from example
+    private:
+        int b;
+    public:
+        int getB() { return b; }
+};
+#+END_SRC
+
+* Polymorphism
+- Now we can create =child= objects with all the properties of an =example= object
+- This means we can safely cast a =child= object to an =example= object
+- TODO I'm not a fan of this slide layout - ideas?
+#+BEGIN_SRC c
+child c;
+example& e = dynamic_cast<example&>(c);
+#+END_SRC
+#+BEGIN_SRC c
+// The inverse will not work
+example e;
+child& c = dynamic_cast<child&>(e);
+#+END_SRC
+#+BEGIN_SRC c
+example.cpp: In function ‘int main()’:
+example.cpp:20:38: error: cannot dynamic_cast ‘e’ (of type ‘class example’) to type ‘class child&’
+(source type is not polymorphic)
+     child& c = dynamic_cast<child&>(e);
+                                      ^
 #+END_SRC
 
 * Questions?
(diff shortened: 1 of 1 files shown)


JasonGibson274 (commented):
I would have some examples of polymorphism, inheritance and how/why we group methods and variables together in different classes.

slides/week5/week5.org
```diff
@@ -0,0 +1,80 @@
...
+#+BEGIN_SRC c
+class example {
+    private:         // private is also implied if you don't specify visibility
+        int a;
+    public:
+        int getA() { return a; }
+};
+#+END_SRC
+
+* Why bother?
```
jgkamat, line 30: Maybe you could discuss code reuse a bit here

slides/week5/week5.org
```diff
@@ -0,0 +1,80 @@
...
+
+* What are classes?
+- Definitions for objects, i.e. what data and methods they have
+- An object is called an *instance* of a class
+
+* A simple example
+#+BEGIN_NOTES
+A class can have some data and some methods, which can either be public or private. Private means only visible to the class itself, public means visible to any other file. Drop to a terminal and use/modify this class.
+#+END_NOTES
+#+BEGIN_SRC c
```
jgkamat: These should be `C++` not `c` (they produce compiler errors, and don't recognize C++ only keywords in highlighting).

slides/week5/week5.org
```diff
@@ -0,0 +1,80 @@
...
+
+* Inheritance
+- Classes can inherit members from other classes
+#+BEGIN_SRC c
+class child: public example {
+    // we get `a` and `getA()` from example
+    private:
+        int b;
+    public:
+        int getB() { return b; }
```
jgkamat, line 60: maybe show a `getAPlusB` method to show how it actually is working

slides/week5/week5.org
```diff
@@ -0,0 +1,80 @@
...
+};
+#+END_SRC
+
+* Polymorphism
+- Now we can create =child= objects with all the properties of an =example= object
+- This means we can safely cast a =child= object to an =example= object
+- TODO I'm not a fan of this slide layout - ideas?
+#+BEGIN_SRC c
```
jgkamat: Maybe:
```
- Now we can create =child= objects with all the properties of an =example= object
- This means we can safely cast a =child= object to an =example= object
  #+BEGIN_SRC C++
    child c;
    example& e = dynamic_cast<example&>(c);
  #+END_SRC
- But we can't go the other way
  #+BEGIN_SRC C++
    example e;
    child& c = dynamic_cast<child&>(e);
  #+END_SRC

  #+BEGIN_EXAMPLE
    example.cpp: In function ‘int main()’:
    example.cpp:20:38: error: cannot dynamic_cast ‘e’ (of type ‘class example’) to type
    ‘class child&’  (source type is not polymorphic)
         child& c = dynamic_cast<child&>(e);
  #+END_EXAMPLE
```

jgkamat (changes requested):
Overall looking pretty good, but needs a few changes.

slides/week5/week5.org
kberzinch: Ooh, I like that

slides/week5/week5.org
kberzinch: I tried that but the bullet point in the middle didn't line up with the rest of them and it looked kind of dumb

jgkamat: Hmm, that's weird. This is what I get, fwiw.

2017-07-31-212722_1920x1080_scrot

ebarnette:
@kberzinch When you talk about what a class is, one of the best metaphors to use is that classes are like instructions/blueprints sent to a factory to build a lot of the same thing. "Definitions for objects" is a little vague. I feel like you can flesh out that slide more.

slides/week5/week5.org
ebarnette: Major plus 1 on this. The slide on inheritance is hella vague.

slides/week5/week5.org
ebarnette: +1

slides/week5/week5.org
kberzinch: Added

slides/week5/week5.org
kberzinch: I added a bullet point but I'm not a fan of the phrasing. Suggestions appreciated.

slides/week5/week5.org
kberzinch: OHHH I see what you did there. Fixed and pushed!

slides/week5/week5.org
jgkamat: Maybe: "Additional functionality can be added with minimal extra code using inheritance"?

jgkamat (changes requested):
The content here looks mostly good to me, but there needs to be more content explaining what an object and what a class is (and what's the difference). I think the existing example jumps into code too early.

slides/week5/week5.org
kberzinch: I like it - added

jgkamat (dismissed):
Overall, I think that the content of the slides looks good to me, but it still feels confusing for a newcomer. This topic is really hard though, so we should just focus on answering questions imo.

slides/week5/week5.org
```diff
@@ -0,0 +1,85 @@
...
+* What are we doing today?
+- Classes in C++
+- Handling merge conflicts in =git=
+
+* What are classes?
+- Definitions for objects, i.e. what data and methods they have
+ - In other words, a blueprint for how to create an object
+- An object is called an *instance* of a class
+ - An object has all the functions and data structures defined in its class
+ - It doesn't share any data with other objects
```
JasonGibson274: I feel like this is misleading. Objects can share data with other objects. Perhaps something like "Objects of the same class share the same schema of data but the actual values stored are independent instance to instance"

JasonGibson274 (changes requested):
Overall looks good to me.

JasonGibson274:
Actually we are missing the virtual keyword explanation.

kberzinch:
Added @JasonGibson274

JasonGibson274 (dismissed):

jgkamat (dismissed):

slides/week5/week5.org
```diff
@@ -0,0 +1,91 @@
...
+- Software projects are easier to understand when related data and functions are grouped together
+- Abstracting data behind getters/setters allows you to validate inputs from other parts of your application
+- Abstracting complex tasks into class methods enables simpler, easier to read high-level code
+- Classes can be extended to add functionality with minimal extra code using inheritance
+
+* Static members
+#+BEGIN_NOTES
+ Drop to a terminal and use this class.
+#+END_NOTES
+- For when you want data or functions to be part of your class, but they don't need to be "attached" to an instance of the class
```
joshhting, line 40: This wording seems a little more confusing than explicitly saying that all instances of a class read from and write to the same static variable

kberzinch: Added.

joshhting (approved):

PR #42, opened 2017-08-04
Week five example code

JasonGibson274:
I tried to keep the classes simple. I can add another class at the end with examples of constructor chaining, use of this, and other more advanced topics but I just wanted something rather simple.


code/week5/calculator.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+int Calculator::subtract(int a, int b) {
+	return a - b;
+}
+
+double Calculator::subtract(double a, double b) {
+	return a - b;
+}
+
+int Calculator:: multiply(int a, int b) {
```
kberzinch: Remove extra space here

kberzinch (changes requested):
I like it, just found one small typo.

code/week5/week5.cpp
```diff
@@ -0,0 +1,61 @@
...
+	judy_bunny_object->do_something();
+	judy_bunny_object->do_something();
+	judy_bunny_object->print_ticket_numbers();
+	cout << "there are " << Bunny::get_meter_maid_number() << " meter maids" << endl;
+	unique_ptr<Bunny> another_bunny = make_unique<Bunny>("Judy", 10);
+	cout << "there are " << Bunny::get_meter_maid_number() << " meter maids" << endl;
+	cout << "\n";
+
+	/************* DYNAMIC CASTING *************/
+	Animal* casted_predator = dynamic_cast<Animal*>(make_unique<Predator>("John").get());
```
barulicm: I don't want to be mean....but this code........it hurts...........
But, in all seriousness, this code actually won't (or at least shouldn't) work.
make_unique(...) gives you a unique_ptr, with ownership of a dynamically allocated Predator object. You then get the raw pointer to that object, pass it through a dynamic cast, and save it off to a pointer to an Animal object.
But wait! That temporary unique_ptr we made earlier is no longer needed so the computer will happily destroy it for us. And, the unique_ptr being the responsible memory owner that it is, will delete the Predator object it allocated.
Every line after this is at best undefined behavior, but most likely will just segfault.
It is generally bad practice to mix and match the old- and smart-style pointer paradigms because you expose yourself to all of the same problems as just using old-style pointers with a misleading sense of comfort given by the presence of smart pointer types and methods.
A better example for this functionality would be to use shared_ptr and dynamic_pointer_cast:
```
auto casted_predator = dynamic_pointer_cast<Animal*>(make_shared<Predator>("John"));
```
Shared pointers share ownership, so the fact that the temporary shared_ptr from make_shared gets deleted here doesn't matter, as casted_predator still participates in owning the memory.

code/week5/animal.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <iostream>
+
+class Animal {
+	std::string name;
```
barulicm: I, personally, am not a fan of the implied access modifiers (the fact that the members declared before "public:" here default to private). At the end of the day, this is just a style choice, but I prefer to explicitly state "private:" so intentions are clear. This also allows you to move private member declarations to the end of the class declaration, moving your public interface at the top. This is nice as most people reading your header will be primarily interested in the public interface of your class.

code/week5/animal.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <iostream>
+
+class Animal {
```
barulicm, line 5: Base classes in C++ which you intend to be used polymorphically must declare virtual destructors (even if they are just default). All classes here which are derived from should have a `~T() = default;` line in their declaration.

code/week5/bunny.h
```diff
@@ -0,0 +1,28 @@
...
+	int ticket_num;
+public:
+	// constructor
+	Bunny(std::string name, int required_tickets);
+
+	// overwrites the do_something method from animal and will be called
+	// even if a Bunny is referenced as a Animal
+	// override annotation has the compiler check that the method in overriden
+	// correctly i.e. this prototype matches one in Animal of the same name.
+	virtual void do_something() override;
```
barulicm: The virtual keyword is redundant and generally omitted in the presence of the override keyword.

code/week5/bunny.h
```diff
@@ -0,0 +1,28 @@
...
+// subclass of Prey through that a subclass of Animal
+class Bunny : public Prey {
+	static int meter_maid_number;
+	int required_tickets;
+	int ticket_num;
+public:
+	// constructor
+	Bunny(std::string name, int required_tickets);
+
+	// overwrites the do_something method from animal and will be called
```
barulicm: override, not overwrite. 
It's levio-saaahhhh

code/week5/week5.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
barulicm: Is there a reason these includes are in a separate file and not just at the top of main.cpp?

barulicm (changes requested):

code/week5/week5.h
jgkamat: If an additional file (`main2.cpp`) is added which includes `week5.h` and does not `#pragma once` in it's own cpp file, it won't get the `#pragma once`, which can be a gotchya.

barulicm: No, I meant, why does week5.h exist as a file? The only place I see it being included is in main.cpp, so this list of headers should just live directly in main.cpp, with no need for this extra file.

jgkamat: oh yah, didn't read far enough down, :+1:

code/week5/week5.cpp
jgkamat: (I don't know that much about cpp casting but)

:+1: on above, and:

since this is a dynamic cast, it's probably a good idea to do a check if casted_predator is null before dereferencing it, since otherwise it could just be an implicit upcast (I think that exists). I'm not sure how implicit casting works with shared pointers though.

barulicm: My bad, the above example can just use static_pointer_cast to achieve the desired casting. I don't think any runtime checks would be necessary at that point.

JasonGibson274: Please be as critical as possible. It helps both me and the code.

JasonGibson274: I fixed the line in question and left it in as an example of dynamic casting. I can do something more with it if anyone wants. I just wanted it to as simple as possible.

code/week5/animal.h
JasonGibson274: added destructor for all classes.

barulicm (dismissed):

code/week5/week5.cpp
```diff
@@ -0,0 +1,102 @@
...
+		casted_prey->do_something();
+		casted_prey->say_your_name();
+		delete casted_prey;
+		cout << "\n";
+	}
+
+	Prey* casted_animal = dynamic_cast<Prey*>(new Animal("Fail"));
+	// dynamic casting can fail
+	if(!casted_animal) {
+		cout << "casting failed" << endl;
```
jgkamat: (not entirely sure but)

This feels like a memory leak. If the casting fails, the new animal is created, and dropped in the cast. Since it's never in a smart pointer, the destructor isn't called. I don't think that the destructor is called by the dynamic cast, but I could be wrong.

(also same thing on line 77 as well)

code/week5/week5.cpp
```diff
@@ -0,0 +1,102 @@
...
+		cout << "casting successful" << endl;
+		delete casted_animal;
+	}
+	cout << "\n";
+	
+	{
+		// pointer casts also exist
+		shared_ptr<Animal> casted_predator = static_pointer_cast<Animal>(make_shared<Predator>("John"));
+		casted_predator->say_hello();
+		casted_predator->do_something(); 		casted_predator->say_your_name();
```
jgkamat: Having two code lines on one line is bugging me a bit :smile:

code/week5/week5.cpp
```diff
@@ -0,0 +1,102 @@
...
+		prey_object->say_your_name();
+		prey_object->set_name("Not Emma");
+		prey_object->say_your_name();
+		prey_object->say_hello();
+		prey_object->do_something();
+	}
+	cout << "\n";
+
+	/************* BUNNY *************/
+	{
```
jgkamat: The file above this seems to be tabs, and below this seems to be spaces. :/

code/week5/prey.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include <iostream>
```
jgkamat, line 3: tabs instead of spaces (:cry:)

(actually a bunch of files have this, so I'm going to just leave this one)

[install a plugin](http://editorconfig.org/#download) to avoid worrying about this in the future...

jgkamat (changes requested):
Mostly a bunch of small things, looking very good overall :tada:

code/week5/week5.cpp
barulicm: Woah, totally missed this. This casting will always fail. new Animal(...) creates an Animal object. When you down cast this to a Prey pointer, dynamic_cast will always return nullptr because the pointed-to object is not a Prey object (nor a derived class of Prey).
http://en.cppreference.com/w/cpp/language/dynamic_cast
See item 5 in the Explanation section in the link above.

JasonGibson274: That was intentional. I was showing an example of a casting dialing. I will make the comment more explicit or I can take it out.

jgkamat: I think you should leave it in just to have an example of dynamic casting. I think you should improve the comment and fix the leak (if it exists) though.

code/week5/week5.cpp
```diff
@@ -0,0 +1,106 @@
+/*
+ * to compile
+ * g++ -std=c++14 -o week5.out *.cpp
+ */
+#include <iostream>
```
jgkamat, line 5: This isn't compiling for me without

```
#include <memory>
```
to get the defs for unique_ptr and such

code/week5/week5.cpp
```diff
@@ -0,0 +1,106 @@
...
+    cout << "\n";
+
+    /************* DYNAMIC CASTING *************/
+    // this can also be done with static cast since types are known at compile time
+    // and have zero chance of being anything else
+    Animal* casted_prey = dynamic_cast<Animal*>(new Predator("Liam"));
+    // this checks if the dynamic casting failed
+    // this checks for a null pointer i.e. pointer == 0
+    if(!casted_prey) {
+        cout << "casting failed" << endl;
```
jgkamat: If (for some reason) the dynamic cast fails (it won't in this case, because it's hard coded, and an upcast) this will be a leak, so it's probably best to account for both cases (even though practically only one will run in this case).

jgkamat (changes requested):
Just a few more nitpicks, I think everything else LGTM though :D

code/week5/week5.cpp
JasonGibson274: I actually do not know how mine is compiling.

kberzinch (changes requested):
You committed a compiled binary

code/week5/week5.cpp
jgkamat: some compilers auto-include "useful" libraries and print a warning, for example, gcc will auto-include iostream with a warning, so that might be it. :P

jgkamat:
(we should probably add `*.out` to the root .gitignore). I suspect a lot of the students will have the same problem :smile:

kberzinch:
I second that!

kberzinch (approved):

JasonGibson274:
Interestingly enough I added that but it was during week 2 so it is in that branch but not in master.

jgkamat (approved):
LGTM :+1:

PR #43, opened 2017-08-05
added .out files to gitignore

JasonGibson274:
Added .out to gitignore


jgkamat (approved):
LGTM. We need to cherry pick this to all the branches though, I think. (unless we were fine with having it in only at master)

kberzinch (approved):

PR #44, opened 2017-08-07
Update weekFour README

joshhting:


jgkamat (approved):
Content LGTM, but this is probably going to cause conflicts later on (since this isn't a cherry pick or a merge). We'll handle that later though (probably) :smile:

PR #45, opened 2017-08-08
Week six topic slides

chsahit:


slides/week6/week6.org
```diff
@@ -0,0 +1,57 @@
...
+  This means each row of the image is appended onto the end of the last. To
+  iterate through a Mat you just get a pointer to the beginning of the first
+  row and keep track of your row number by how far you've traversed.
+
+* Color Images
+- Color images don't embed the color of a pixel in one element. Often, you'll
+  find each pixel represented in BGR (Blue component, Green Component, Red
+  Component) form. So now, each row of a color image is 3 times as long as a
+  row of a black and white image.
+[[https://docs.opencv.org/2.4/_images/math/b6df115410caafea291ceb011f19cc4a19ae6c2c.png]]
```
jgkamat: Previously, we were prepending `file:` to images to work around a bug in the org markdown exporter.

This has [now been fixed](http://repo.or.cz/org-mode.git/commit/f3d7284ebd0c6e8248f7c3c444937270402b0ba6), and both methods should work.

Luckily, a [new snapshot was released recently](https://elpa.gnu.org/packages/org.html), so this format should be working now, since our script pulls from the latest snapshot.

So tl;dr: the other image links look different, will continue to work, but this new format should work and is preferred, because it integrates better with the org syntax checker (let me know if anything breaks).

Also, unrelated (and the reason I made this comment) is that this link seems to be broken:
https://docs.opencv.org/2.4/_images/math/b6df115410caafea291ceb011f19cc4a19ae6c2c.png

slides/week6/week6.org
```diff
@@ -0,0 +1,57 @@
...
+- We have the knowledge in C++ to describe the logic we might want a robot to
+  have. But we need to be able to make sense of what the robot sees and
+  classify it before we can act on this logic.
+- Cue OpenCV, an open source computer vision library with bindings for C++
+  (and a few other languages)
+- I guess our ability to see has been ++'d
+
+* What does an Image look like to your computer?
+- OpenCV stores images in an object called a /Mat/
+- A Mat is matrix with rows and columns! In a grayscale image. each element
```
jgkamat: `In a grayscale image<.>` You probably want a comma.

jgkamat: Also, `A Mat is matrix with rows and columns`. 

(If I'm understanding `Mat`'s correctly)

If I know what a matrix is: This information is redundant.
If I don't know what a matrix is: This leaves me confused (ie: what is a matrix)

I would suggest just saying an array of rows and columns (we covered arrays before), or just saying it's a table (or similar), or just saying it's a matrix.

slides/week6/week6.org
```diff
@@ -0,0 +1,57 @@
...
+* HSV Images
+- Each Pixel in a color image has a hue, a saturation, and a luminosity. 
+- The nice thing about looking at colors from this perspective is that we're
+  more tolerant to changes in lighting conditions making the same color look
+  different.
+[[https://image.slidesharecdn.com/01presentationhuehistograms-150707215651-lva1-app6892/95/about-perception-and-hue-histograms-in-hsv-space-5-638.jpg]]
+
+* Finding the blue in an image
+#+BEGIN_SRC c++
+Mat findBlue(const Mat& frameBGR) {
```
jgkamat: I'm kind of lost here, I'm not exactly sure what's going on (I haven't done CV before, so that's expected I guess).

Could you add either speaker notes or comments to this bock?

eg:

- Why are we blurring at all if we want to find the blue?
- What exactly is "Finding the blue" are you finding blue overall, or blue in a region, or...
- what is "CVT" in cvtColor?

jgkamat (changes requested):
Overall comment is that I'm a little bit lost at the end. I'm  not sure if the examples we'll be going through is in this presentation or not though.

slides/week6/week6.org
jgkamat: oh it just seems like docs.opencv.org is down, probably this will be fine when that comes back up.

kberzinch: I get a certificate error.

jgkamat: its both a cert error and something else wrong (seems like a default page). Maybe they're working on migration or something.

2017-08-08-174301_880x580_scrot

jgkamat: oh I got it, docs.opencv.org dosen't support https, but still has an endpoint:

http://docs.opencv.org/2.4/_images/math/b6df115410caafea291ceb011f19cc4a19ae6c2c.png

Because our website is https, we still have to make sure all images are https (or browsers might try to block them), I would suggest re-uploading it to an image host.

JasonGibson274:
Also can we have slides on how and where to install OpenCV?

slides/week6/week6.org
```diff
@@ -0,0 +1,57 @@
...
+- Color images don't embed the color of a pixel in one element. Often, you'll
+  find each pixel represented in BGR (Blue component, Green Component, Red
+  Component) form. So now, each row of a color image is 3 times as long as a
+  row of a black and white image.
+[[https://docs.opencv.org/2.4/_images/math/b6df115410caafea291ceb011f19cc4a19ae6c2c.png]]
+- Images don't have to be stored in just BGR format!
+
+* HSV Images
+- Each Pixel in a color image has a hue, a saturation, and a luminosity. 
+- The nice thing about looking at colors from this perspective is that we're
```
joshhting: I would be even more explicit with describing which components of rgb vs hsv are affected/unaffected by lighting conditions

slides/week6/week6.org
```diff
@@ -0,0 +1,57 @@
...
+#+AUTHOR: Sahit Chintalapudi 
+#+EMAIL: schintalapudi@gatech.edu
+
+* Computer Vision
+- We have the knowledge in C++ to describe the logic we might want a robot to
+  have. But we need to be able to make sense of what the robot sees and
+  classify it before we can act on this logic.
+- Cue OpenCV, an open source computer vision library with bindings for C++
+  (and a few other languages)
+- I guess our ability to see has been ++'d
```
joshhting, line 11: +1

joshhting (changes requested):

slides/week6/week6.org
chsahit: So I'll go ahead and add some speaker notes, but if you want the fully commented version see 
https://github.com/RoboJackets/software-training/blob/weekSixExampleCode/code/week6/findBlue.cpp 
to get read up on what's happening.
To answer your three listed questions, blurring helps remove noisey pixels by smoothing over the image. "Finding the blue" is taking an entire image and identifiying all parts of the image that are the color blue.  The cvt in cvtColor is convert because you're converting from RGB to HSV color space.

jgkamat: Ah, that makes a lot more sense now, thanks for the follow up. I have a few more questions though:

* The low value that you're picking `HSV(78, 50, 70)` seems 'green' to me, [is that normal](https://duckduckgo.com/?q=HSV(78%2C50%2C70)&kc=1&k8=a6aaab&k1=1&kx=909396&ky=2e2e2e&kaa=B68800&kp=1&k9=ce4139&kj=262626&kam=osm&k5=1&kaj=m&kn=-1&k7=262626&kao=1&kae=d&kv=1&kak=-1&ia=answer).
* Can you link some source on what a CVT is? I'm only finding a super fancy transmission for cars when I search or do an acronym lookup.

I'm fine with leaving the comments and notes out personally, since the comments are in the code.

chsahit: so what i mean by 

> The cvt in cvtColor is convert because ....

is that "cvtColor" is literally shorthand for "convertColor". Not sure if I said that clearly before. If I did, and you were asking for source to prove that the word cvt was short for the word convert than the documentation on the method might be all that I can do for you and that isn't very explicit. 
http://docs.opencv.org/2.4/modules/imgproc/doc/miscellaneous_transformations.html#cvtcolor

As for why the low HSV seems green, I see the issue. It was in our best interests to err on the wide side for our color bounds for blue because not a lot else on the track was blue or neighboring colors (that green that you see). Any noisy pixels would be filtered out by later postprocessing. Those bounds were the actual ones we were going to use for IARRC this year where only the finish line was blue.

chsahit:
@sastorer @joshhting @jgkamat I made the requested changes, I think this needs another review so it can  get merged in?

slides/week6/week6.org
```diff
@@ -0,0 +1,65 @@
...
+  This means each row of the image is appended onto the end of the last. To
+  iterate through a Mat you just get a pointer to the beginning of the first
+  row and keep track of your row number by how far you've traversed.
+
+* Color Images
+- Color images don't embed the color of a pixel in one element. Often, you'll
+  find each pixel represented in BGR (Blue component, Green Component, Red
+  Component) form. So now, each row of a color image is 3 times as long as a
+  row of a black and white image.
+[[https://i.imgur.com/QlokNTv.png]]
```
jgkamat: If this is indented to the same level as the list (so it's part of the list) the bullet points will become aligned. Discontinuing a bulleted list and then immediately resuming it is confusing org reveal.

```
@@ -28,11 +28,13 @@
   find each pixel represented in BGR (Blue component, Green Component, Red
   Component) form. So now, each row of a color image is 3 times as long as a
   row of a black and white image.
-[[https://i.imgur.com/QlokNTv.png]]
+
+  [[https://i.imgur.com/QlokNTv.png]]
+
 - Images don't have to be stored in just BGR format!
```

slides/week6/week6.org
```diff
@@ -0,0 +1,65 @@
+#+TITLE: Week 6
+#+AUTHOR: Sahit Chintalapudi
```
jgkamat: There is a bunch of trailing whitespace in this file (this line is an example).

slides/week6/week6.org
```diff
@@ -0,0 +1,65 @@
...
+      cvtColor(frameBlurred, frameHSV, CV_BGR2HSV); 
+      Mat output_blue = Mat::zeros(frameHSV.height, frameHSV.width, CV_8U);
+      inRange(frameHSV, blue_low, blue_high, output_blue);
+      erode(output_blue, output_blue, erosion_kernel_blue);
+      return output_blue;
+}
+#+END_SRC
+
+* Roboracing Computer Vision
+[[https://github.com/RoboJackets/roboracing-software/blob/master/iarrc/src/color_detector/color_detector.cpp]]
```
jgkamat: Having this be a plain link means that it takes up a lot of space and no one will copy it. If you want people to copy it from the slide deck please put a shortened link as the link title.

slides/week6/week6.org
```diff
@@ -0,0 +1,65 @@
...
+[[https://i.imgur.com/QlokNTv.png]]
+- Images don't have to be stored in just BGR format!
+
+* HSV Images
+- Each Pixel in a color image has a hue, a saturation, and a luminosity. 
+- Even though our cameras read in images with RGB, converting them to HSV is
+  easy with OpenCV
+[[https://image.slidesharecdn.com/01presentationhuehistograms-150707215651-lva1-app6892/95/about-perception-and-hue-histograms-in-hsv-space-5-638.jpg]]
+
+* Why do we use HSV
```
jgkamat, line 40: This is a really good slide, like it a lot! :)

jgkamat (changes requested):
Overall, this is looking a lot better, nice! I think we'll be done in one more round :smile:

slides/week6/week6.org
chsahit: I'm not sure if the image looks good with a little bullet next to it, but it's all lined up so that helps readability a lot. Handled your other valid points so let me know what you think, thanks for the feedback!

jgkamat (approved):
Overall LGTM. It feels a little bit short (esp compared to other topic slides), but the content is good. 

Maybe you could add a few more slides discussing (but not implementing) more advanced CV concepts (blobbing?), but I'm not sure how we're planning on managing time this meeting.

slides/week6/week6.org
jgkamat: Also, if you would like to remove the bullet, you can. Just indent it 2 spaces (to the rest of the bulleted list) without the actual bullet char.

joshhting (approved):

PR #47, opened 2017-08-11
Week six example code

JasonGibson274:
@chsahit I made the findBlue.cpp code compile and it seems to work, double check me to make sure I did not break anything without realizing it. I have not checked this code working on virtualbox yet. I will update the label when I do.


code/week6/findBlue.cpp
```diff
@@ -0,0 +1,82 @@
...
+        if(!cameraFrame.empty()) {
+            // creates a window to display the actual camera image
+            imshow("camera", cameraFrame);
+            Mat blue = findBlue(cameraFrame);
+
+            // creates a window to display the blue parts of the image
+            imshow("blue parts", blue);
+
+            // waits and displays the image for 1 millisecond
+            waitKey(1);
```
jgkamat: Right now, closing the window does not quit the program, is it possible to do that? It's a little annoying to close the way it's set up currently.

JasonGibson274: now it will close if any key is pressed but closing the window will do nothing. Is that acceptable?

jgkamat: yup, that sounds good to me, at least. :smile:

JasonGibson274:
works on VM provided that extensions pack is installed.

code/week6/main.cpp
```diff
@@ -0,0 +1,58 @@
+#include <opencv2/opencv.hpp>
```
jgkamat, line 1: I think this is the first time we're using Cmake in the course, can you add a list of commands to run to this file (or add a makefile?)

```
mkdir build
cd build
cmake ../
make -j
./findBlue
```

code/week6/findBlue.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+int main() {
+    // sets up the input stream to the default video device
+    // /dev/video0 must exist
+    VideoCapture stream1(0);
+
+    // this checks to see if a key has been pressed and then shows the image for
+    // one millisecond
+    while (waitKey(1) == 255) {
```
jgkamat: For me this causes the image to never come up at all. You probably want:

`waitKey(1) == -1`.

I also take back this being an OK solution, since you might want to press a key and have the program keep running (for example, my window dragging workflow needs to have a key pressed, so I'm unable to manipulate the window).

Perhaps, you could use:
```
(char)waitKey(1) != 'q'
```
or similar.

jgkamat (changes requested):
Just a few more usability fixes

code/week6/main.cpp
barulicm: No, we'll be using CMake for all of the hardware application demos for every week. They'll be familiar with this workflow by this point in the semester.

code/week6/findBlue.cpp
barulicm: While the suggested solution is a better way to implement this functionality in general, boo to non-standard window managers. Specifically, don't worry too much about them, since none of the students will be using them.

jgkamat: >Specifically, don't worry too much about them, since none of the students will be using them. 

There was one student last year in robocup training (who left unfortunately) who used i3 (and many returning robocup team members, including will, at least partially use i3 or other tiling window managers).

There was an incoming robocup member who used emacs + unity to make his own mouseless workflow (which was what inspired me to make my own).

> boo to non-standard window managers

Assuming you're talking about unity (since it seems to be dominant at rj) unity is [officially deprecated by ubuntu](https://insights.ubuntu.com/2017/04/05/growing-ubuntu-for-cloud-and-iot-rather-than-phone-and-convergence/) and will be dropped in future ubuntu versions.

While we shouldn't prioritize alternative workflows, we should solve breaking issues as they are discovered, to avoid alienating future members. I would be happy to do this if no one else is willing to, of course. Diversity in all it's forms is important for organizations to have a balanced viewpoint.

JasonGibson274: so interestingly enough my computer returns 255 when nothing is pressed.

jgkamat: Interesting. It's probably something with an implcit cast to unsigned car not happning on my machine.

jgkamat (dismissed):
LGTM, but I don't know too much about opencv, so you probably want another opinion.

I tested this on opencv3 and 2, and it seems to work on both.

ebarnette:
@barulicm can you give this a review before it's merged?

JasonGibson274:
just missing roboracing video @chsahit

JasonGibson274:
@barulicm Does this look better? I slowed the video down a lot so that you can actually see what was happening by comparing the two images but I can speed it up if that is preferred.

code/week6/findBlue.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+      // kernel that is an ellipse of size 11X11 pixels
+      Mat erosion_kernel_blue = getStructuringElement(MORPH_ELLIPSE, Size(11, 11));
+
+      //applying a GaussianBlur makes the image "fuzzier". This smoothes out
+      //noisy pixels
+      Mat frameBlurred;
+      GaussianBlur(frameBGR, frameBlurred, Size{7,7}, 0);
+
+      //We shift the mat which is encodes as an RGB image into an HSV image
```
barulicm: *encoded (not encodeS)

code/week6/findBlue.cpp
```diff
@@ -0,0 +1,90 @@
...
+      Mat frameHSV;
+      cvtColor(frameBlurred, frameHSV, CV_BGR2HSV);
+
+      //We initialize an empty Mat to store all the pixels that we identify
+      //as blue. CV_8U is the image depth: each pixel is only one bit
```
barulicm: each pixel is only one *byte*

code/week6/findBlue.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+      //for every pixel in our frame that we define as blue, write a white
+      //pixel to the output_blue Mat
+      inRange(frameHSV, blue_low, blue_high, output_blue);
+
+      //apply an erode operation onto the filtered image. This helps us
+      //remove specks of blue that show up due to noise or glare
+      erode(output_blue, output_blue, erosion_kernel_blue);
+
+      //the logic to publish the result of the filtering in ROS would go here
```
barulicm: No real need to mention ROS in this demo

code/week6/findBlue.cpp
```diff
@@ -0,0 +1,90 @@
...
+    return output_blue;
+}
+
+int main() {
+    // sets up the input stream to the default video device
```
barulicm: This loads from a video, not the default video device.

code/week6/findBlue.cpp
```diff
@@ -0,0 +1,90 @@
...
+    // the file must be in the local directory
+    VideoCapture video("findBlueDemo.mp4");
+
+    if(!video.isOpened()) {
+        std::cout << "Error opening video stream or file" << std::endl;
```
barulicm: Do we want to use `std::cerr`?

code/week6/findBlue.cpp
```diff
@@ -0,0 +1,90 @@
...
+        std::cout << "Verify that findBlueDemo.mp4 is in the local directory" << std::endl;
+        return -1;
+    }
+
+    // this checks to see if a key has been pressed and then shows the image for
+    // 300 milisecond so you can see what is happening
+    while (static_cast<char>(waitKey(300)) != 'q') {
+        Mat cameraFrame;
```
barulicm: In any real-world code this would be declared outside of the loop for performance reasons

code/week6/main.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <string.h>
+#include <stdlib.h>
+
+int main() {
+    using namespace std;
```
barulicm: Why is this inside the main method? Also we should be consistent about which namespaces we include. The findBlue demo uses cv and makes std explicit, this demo does the opposite.
I recommend not including the cv namespace so it is obvious which functions are from OpenCV as opposed to standard c++.

code/week6/main.cpp
```diff
@@ -0,0 +1,58 @@
...
+     * C3 means that the color channel has three entires (BGR)
+    */
+    my_image = cv::Mat(2, 2, CV_8UC3, cv::Scalar(0,0,255));
+
+    // a sinlge pixel is made up of 3 8 bit values
+    // Blue, Green, Red
+    cout << "my_image = \n" << my_image << endl;
+
+    // iterate over an image
+    for(auto it = my_image.begin<cv::Vec3b>(); it != my_image.end<cv::Vec3b>(); it++) {
```
barulicm: This is not the fastest way to iterate over an image in OpenCV. Do we care?

code/week6/main.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+        // applies to laplacian kernel, a common technique for edge detection
+        cv::Laplacian(image, image, image.depth(), 3, 1, 0, cv::BORDER_DEFAULT);
+
+        // creates a window and shows the image until a key is pressed
+        cv::namedWindow("Jaymii", cv::WINDOW_AUTOSIZE);
+        cv::imshow("Jaymii", image);
+        cv::waitKey(0);
+    }
+}
```
barulicm: *main()* reaches the end without returning anything.

barulicm (changes requested):

code/week6/main.cpp
JasonGibson274: more for my own benefit here. Isn't return 0 implicitly added anyway so why is it worth making it explicit?

code/week6/main.cpp
JasonGibson274: do we want an example of iterating over a Mat or should this just be taken out?

code/week6/findBlue.cpp
JasonGibson274: Can you explain the performance issue here?

code/week6/main.cpp
barulicm: That's up to someone else. It's a useful thing to know, but you can also do a lot without using it.

JasonGibson274:
@barulicm final checks

code/week6/findBlue.cpp
```diff
@@ -0,0 +1,88 @@
+#include <unistd.h>
```
barulicm: Why do we need unisdt.h here?

code/week6/main.cpp
```diff
@@ -0,0 +1,60 @@
+#include <opencv2/opencv.hpp>
+#include <string.h>
```
barulicm: Use current standard headers ie <string>

code/week6/main.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <stdlib.h>
```
barulicm: Why do we need stdlib.h here?

code/week6/main.cpp
barulicm: This comment still stands

code/week6/main.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+    // check that the iamge was loaded
+    if(!image.empty()) {
+        cv::Vec3b image_vector = image.at<cv::Vec3b>(154,160);
+        cout << "Blue = " << to_string(image_vector[0]) << endl;
+        cout << "Green = " << to_string(image_vector[1]) << endl;
+        cout << "Red = " << to_string(image_vector[2]) << endl;
+
+        // converts image to grey scale
+        cvtColor(image, image, cv::COLOR_BGR2GRAY);
```
barulicm: Noticing a lack of cv:: here, does this compile?

barulicm (changes requested):
Some of my comments from my last review were not addressed.
Be careful with what appear to be includes suggested by your IDE. C++ standard headers like <string> and <iostream> are preferred over C or system-specific headers.

code/week6/main.cpp
JasonGibson274: This actually does compile on my machine but I will add it explicitly.

code/week6/main.cpp
JasonGibson274: I added something along the lines of this http://docs.opencv.org/2.4/doc/tutorials/core/how_to_scan_images/how_to_scan_images.html#howtoscanimagesopencv

JasonGibson274:
@barulicm sorry I missed those changes in mentioned earlier. If there is a faster/better way to iterate over an image let me know that is what I found on the opencv docs.

barulicm (approved):

PR #48, opened 2017-08-15
Week seven example code

JasonGibson274:
the print_data function still needs some work (restricting types) and I can add more examples of using data structures. Also any changes that come from the topic slides.

### DIFF
--- a/code/week7/generic_class.h
+++ b/code/week7/generic_class.h
@@ -9,7 +9,7 @@ public:
     };
     ~Example() {};
 
-    const primitive_type getVar() {
+    primitive_type getVar() const {
         return var;
     }
 
--- a/code/week7/generic_class.h
+++ b/code/week7/generic_class.h
@@ -1,0 +1,16 @@
+#include <iostream>
+
+template <class primitive_type> class Example {
+private:
+    primitive_type var;
+public:
+    Example(primitive_type var) : var(var) {
+        std::cout << "creating object with var = " << var << std::endl;
+    };
+    ~Example() {};
+
+    primitive_type getVar() {
+        return var;
+    }
+
+};
--- a/code/week7/generic_class.h
+++ b/code/week7/generic_class.h
@@ -9,7 +9,7 @@ public:
     };
     ~Example() {};
 
-    primitive_type getVar() {
+    const primitive_type getVar() {
         return var;
     }
 
--- a/code/week7/main.cpp
+++ b/code/week7/main.cpp
@@ -128,15 +128,15 @@ int main() {
 
     // can insert elements into set
     ordered_set.insert("2");
-    print_data(ordered_set.begin(), ordered_set.end(), "ordered_set");
+    print_data(ordered_set.begin(), ordered_set.end(), "ordered_set first time");
 
     // If the same element is inserted TODO
     ordered_set.insert("1");
-    print_data(ordered_set.begin(), ordered_set.end(), "ordered_set");
+    print_data(ordered_set.begin(), ordered_set.end(), "ordered_set second time");
 
     // elements can also be erased
     ordered_set.erase("1");
-    print_data(ordered_set.begin(), ordered_set.end(), "ordered_set");
+    print_data(ordered_set.begin(), ordered_set.end(), "ordered_set third time");
 
     /*************** UNORDERED SETS ***************/
     // http://www.cplusplus.com/reference/unordered_set/unordered_set/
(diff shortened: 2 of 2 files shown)


code/week7/main.cpp
```diff
@@ -0,0 +1,186 @@
...
+    // set<TYPE> name
+    set<string> ordered_set {"1", "5", "6"};
+
+    // can insert elements into set
+    ordered_set.insert("2");
+    print_data(ordered_set.begin(), ordered_set.end(), "ordered_set");
+
+    // If the same element is inserted TODO
+    ordered_set.insert("1");
+    print_data(ordered_set.begin(), ordered_set.end(), "ordered_set");
```
jgkamat: It would be a lot easier to follow in the output if the label changes on the three lines.

code/week7/main.cpp
```diff
@@ -0,0 +1,186 @@
...
+    // init
+    // set<TYPE> name
+    unordered_set<string> unordered_set {"20","19","18","1","16"};
+
+    // can insert elements into set
+    unordered_set.insert("21");
+    // order is not maintained here
+    print_data(unordered_set.begin(), unordered_set.end(), "unordered_set");
+
+    /*************** MAPS ***************/
```
jgkamat, line 154: The maps don't show off the unordered vs plain distinction, since we only add one element. Not sure if we should add it though.

code/week7/generic_class.h
```diff
@@ -0,0 +1,16 @@
...
+template <class primitive_type> class Example {
+private:
+    primitive_type var;
+public:
+    Example(primitive_type var) : var(var) {
+        std::cout << "creating object with var = " << var << std::endl;
+    };
+    ~Example() {};
+
+    primitive_type getVar() {
```
jgkamat: I think it's best practice to mark this function as const, but I don't know if we want to make it easy to read or not at this point.

code/week7/main.cpp
```diff
@@ -0,0 +1,186 @@
...
+    cout << "at int 53 value = " << string_map.at(53) << endl;
+
+    /*************** TEMPLATES ***************/
+    cout << "\n\n";
+    cout << "the max of 7,8 is = " << getMax<int>(7, 8) << endl;
+    cout << "the max of 7.0, 8.0 is = " << getMax<double>(7.0, 8.0) << endl;
+    cout << "the min of 7, 8.0 is = " << getMin<int, double>(7, 8.0) << endl;
+
+    unique_ptr<Example<string>> example_str1 = make_unique<Example<string>>("storing strings here");
+    unique_ptr<Example<int>> exmaple2 = make_unique<Example<int>>(2);
```
jgkamat: You never actually read the data from this level of the program, maybe adding that is a good idea.

jgkamat (changes requested):
Overall, LGTM, just a bunch of minor nitpicks, you can ignore them if you don't think they're good ideas, they're pretty questionable.

code/week7/main.cpp
JasonGibson274: added 3 elements and used iterator

code/week7/generic_class.h
```diff
@@ -0,0 +1,16 @@
...
+template <class primitive_type> class Example {
+private:
+    primitive_type var;
+public:
+    Example(primitive_type var) : var(var) {
+        std::cout << "creating object with var = " << var << std::endl;
+    };
+    ~Example() {};
+
+    const primitive_type getVar() {
```
jgkamat: I actually meant making the method const, not the return type 

ie: `primitive_type getVar() const`. I don't think it matters enough for us to care though.

code/week7/main.cpp
```diff
@@ -0,0 +1,206 @@
...
+        return b;
+    } else {
+        return a;
+    }
+}
+*/
+
+// you can have multiple templates in a function
+template <class Foo, class Bar>
+Bar getMin(Foo a, Bar b) {
```
jgkamat, line 61: This is a little bit... weird. Since we're choosing `Bar` as our return type, we can have a possible loss of precision when using this function and returning a `Foo` (and maybe some casting problems, when using it with non-trivial types).

For example, this patch:

```
From e5fdaf5e8eb5968adc30588f3237bb914b59d24e Mon Sep 17 00:00:00 2001
From: Jay Kamat <jaygkamat@gmail.com>
Date: Sat, 19 Aug 2017 19:24:47 -0400
Subject: [PATCH] Example for breaking return type templates

---
 code/week7/main.cpp | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/code/week7/main.cpp b/code/week7/main.cpp
index 5fbef94..f69e7e6 100644
--- a/code/week7/main.cpp
+++ b/code/week7/main.cpp
@@ -59,7 +59,7 @@ double getMax(double a, double b) {
 
 // you can have multiple templates in a function
 template <class Foo, class Bar>
-Bar getMin(Foo a, Bar b) {
+Foo getMin(Foo a, Bar b) {
     if(a < b) {
         return a;
     } else {
@@ -194,7 +194,7 @@ int main() {
     cout << "\n\n";
     cout << "the max of 7,8 is = " << getMax<int>(7, 8) << endl;
     cout << "the max of 7.0, 8.0 is = " << getMax<double>(7.0, 8.0) << endl;
-    cout << "the min of 7, 8.0 is = " << getMin<int, double>(7, 8.0) << endl;
+    cout << "the min of 7.1, 8.0 is = " << getMin<int, double>(8, 7.1) << endl;
 
     cout << "\n\n";
 
-- 
2.11.0
```

prints out `the min of 7.1, 8.0 is = 7`, which looses the `.1` that was inputed. I don't know of a good way of working around it while keeping the example. If you  have any good ideas, feel free to use them, but I don't have any.

jgkamat (dismissed):
Got a few last notes, but I don't think they matter too much, so I'm approving. :smile:

code/week7/main.cpp
JasonGibson274: I think this just comes down to c++ letting you shoot yourself in the foot if you want to.

jgkamat (approved):

PR #49, opened 2017-08-21
week eight example code

JasonGibson274:
I can always add more usage is necessary

### DIFF
--- a/code/week8/main.cpp
+++ b/code/week8/main.cpp
@@ -28,7 +28,7 @@ void printInt(int i) {
 
 class Parent {
 public:
-    virtual void do_something() {std::cout << "BASE" << std::endl;}
+    virtual void do_something() {std::cout << "PARENT" << std::endl;}
 };
 
 class Child : public Parent {
@@ -38,6 +38,10 @@ public:
 
 void callDoSomething(Parent p) {
     p.do_something();
+}
+
+void callDoSomethingRef(Child c) {
+    c.do_something();
 }
 
 int main() {
@@ -82,7 +86,6 @@ int main() {
     // using lambda
     // [] (TYPE NAME, ...) -> RETURN_TYPE {CODE};
     for_each(example_vec.begin(), example_vec.end(), [](int num) -> void {std::cout << "printing from lambda " << num << std::endl;});
-
     // transform is like for each but it replaces the value with what is returned
     // http://www.cplusplus.com/reference/algorithm/transform/?kw=transform
     transform(example_vec.begin(), example_vec.end(), example_vec.begin(), [](int num) -> int {return num + 2;});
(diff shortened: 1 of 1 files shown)


code/week8/main.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    // transform is like for each but it replaces the value with what is returned
+    // http://www.cplusplus.com/reference/algorithm/transform/?kw=transform
+    transform(example_vec.begin(), example_vec.end(), example_vec.begin(), [](int num) -> int {return num + 2;});
+    printIntVec(example_vec);
+
+    // capture allows us to pass in another var as a constant to the lambda
+    // this makes the variable effectively constant while in the method
+    int constant = 7;
+    transform(example_vec.begin(), example_vec.end(), example_vec.begin(), [constant](int num) -> int {return num + constant;});
```
jgkamat: Not sure if we want to cover this, but I think it would be nice to include a closure example as well, since that's the standard way variable capture work in other languages. 

In cpp, I think it's done by `[&](stuff)`

So maybe 

```
int i = 0;
([&]() -> void {i++;})();
cout << i << endl;
```
Did not try compiling that but that's the idea at least.

jgkamat (commented):
Will come up with an actual review when I get more time.

code/week8/main.cpp
```diff
@@ -0,0 +1,105 @@
...
+    std::cout << "\n\n";
+}
+
+void printInt(int i) {
+    std::cout << "looking at " << i << std::endl;
+}
+
+class Parent {
+public:
+    virtual void do_something() {std::cout << "BASE" << std::endl;}
```
jgkamat: can this print be changed to `parent` or the base class be renamed to `base`, it's a little weird to have multiple names for the same thing.

code/week8/main.cpp
```diff
@@ -0,0 +1,105 @@
...
+    printIntVec(example_vec);
+
+    /******** OBJECT SLICING ********/
+    // object slicing is when a child is assigned to its parent and therefore loses additional features
+    Parent p = Parent();
+    Child c = Child();
+    cout << "parent" << endl;
+    callDoSomething(p);
+    cout << "child" << endl;
+    callDoSomething(c);
```
jgkamat: Just to illustrate how this is related to passing the value, could you add a `callDoSomethingRef` which does use the child function? That way the distinction is more clear.

jgkamat (changes requested):
Overall looks very good, just a few readability nitpicks.

code/week8/main.cpp
```diff
@@ -0,0 +1,110 @@
...
+class Child : public Parent {
+public:
+    virtual void do_something() {std::cout << "CHILD" << std::endl;}
+};
+
+void callDoSomething(Parent p) {
+    p.do_something();
+}
+
+void callDoSomethingRef(Child c) {
```
jgkamat, line 43: I actually meant the header:

`void callDoSomethingRef(Parent& c)`

But we covered that before, so that's not strictly needed.

jgkamat (approved):
LGTM :)

PR #53, opened 2017-08-28
Revert "removed week 7 from master"

JasonGibson274:
This reverts commit 091c8b07d15d87768e58ad880a7b8d75288746b0.

@jgkamat is this the best way to do this?

it was merged into master by accident and now the weekseven branch branched from master after it was deleted. This adds it back in but I wanted to make sure this was the best way to do that.


jgkamat (approved):
This approach seems good to me :)

I can't think of a cleaner way of doing it.

PR #55, opened 2017-08-30
Week1hardware

barulicm:
Opening this for review purposes. Will not be merged until Mac OS X support is finished.


hardware_applications/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+   ```cmake ..```
+   
+4. Build all demos
+
+   ```make```
+
+## How to Rebuild the STSL ##
+
+Binaries and headers for the STSL are provided in the STSL folder. These should work for most circumstances. If you experience problems building the demos with the provided binaries, you may need to rebuild the STSL from source.
```
ebarnette: Expand on the what STSL means and what it is for here. This will be their first introduction to it.

hardware_applications/README.md
```diff
@@ -0,0 +1,71 @@
...
+   
+5. Copy the binaries back to the hardware_applications project.
+   
+   __Windows__
+   
+   Copy _libSTSL.a_ from _build_ to _hardware_applications/STSL/lib/WINDOWS_.
+   
+   __Linux__
+   
+   Copy...
```
ebarnette: Did you mean to actually put the file path here?

hardware_applications/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+   __OS X__
+   
+   Copy...
```
ebarnette: See above comment

hardware_applications/STSL/include/STSL/LinuxUtils.h
```diff
@@ -0,0 +1,20 @@
+#ifndef TRAININGSUPPORTLIBRARY_LINUXUTILS_H
+#define TRAININGSUPPORTLIBRARY_LINUXUTILS_H
+
+#include "OSUtils.h"
+#include <vector>
+
+class LinuxUtils : public OSUtils {
+public:
+    virtual std::string FindRobot() override;
```
ebarnette, line 9: is there a reason you're not using namespaces?

hardware_applications/STSL/include/STSL/OSUtils.h
```diff
@@ -0,0 +1,14 @@
+#ifndef TRAININGSUPPORTLIBRARY_OSUTILS_H
+#define TRAININGSUPPORTLIBRARY_OSUTILS_H
+
+#include <string>
+#include <chrono>
+
+class OSUtils {
+public:
+    virtual std::string FindRobot() = 0;
```
ebarnette: same gripe about namespaces

ebarnette (changes requested):
Looks good. Just a few small requests.

hardware_applications/README.md
barulicm: This is covered by the STSL readme linked to above in this readme.

hardware_applications/README.md
barulicm: At the time this was written, the path did not exist yet.

hardware_applications/README.md
barulicm: See above answer

hardware_applications/STSL/include/STSL/LinuxUtils.h
barulicm: You should never add a `using namespace ...` declaration to a header file, because it would include that namespace in every file which includes that header. This would cause a great deal of annoyance and confusion.

jgkamat: I originally thought the first comment was referring to `using namespace *`, but I think it's referring to 'why can't these files be in their own namespace'?

Also `using namespace *` might in general be a bad idea (causes conflicts): https://github.com/RoboJackets/igvc-software/issues/140

barulicm: @sastorer Can you elaborate on the specifics of what you wanted here?
I'd thought from our offline conversation that you were looking for `using namespace std` here.

barulicm:
@sastorer I believe this is ready to be merged, pending the results of your testing tomorrow morning.

hardware_applications/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+
+include_directories(STSL/include)
+
+if(WIN32)
+    link_directories(STSL/lib/WINDOWS)
+    configure_file(STSL/lib/WINDOWS/libSTSL.dll ${CMAKE_CURRENT_BINARY_DIR}/libSTSL.dll COPYONLY)
+    set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+elseif(UNIX AND NOT APPLE)
+    link_directories(STSL/lib/LINUX)
+    set(CMAKE_CXX_STANDARD_LIBRARIES /usr/lib/x86_64-linux-gnu/libudev.so)
```
jgkamat: hardcoding the paths such as `libudev.so` is really bad practice and makes it so everyone who's not on your specific distribution experiences breakage. The locations of paths can even change between versions of the same operating system, spins of the operating system, different architectures, or versions of the library. For example, it's in `/lib/x86_64-linux-gnu/libudev.so` on my system.

Maybe you could try using [one of the many FindUdev files online](https://duckduckgo.com/?q=findudev+cmake+site%3Agithub.com&kc=1&k8=a6aaab&k1=1&kx=909396&ky=2e2e2e&kaa=B68800&kp=1&k9=ce4139&kj=262626&kam=osm&k5=1&kaj=m&kn=-1&k7=262626&kao=1&kae=d&kv=1&kak=-1&ia=web).

jgkamat (changes requested):

hardware_applications/CMakeLists.txt
barulicm: Yeah I know
Thought I'd fixed this already, but it got lost in all the stuff that happened this week.
Rebooting into Linux to fix this now

hardware_applications/week1/starter_code.cpp
```diff
@@ -0,0 +1,5 @@
+#include <iostream>
+#include <STSL/RJRobot.h>
+
+using namespace std;
+
```
jgkamat, line 5: Since the hard part of this week is getting anything moving, it would be nice to have the most basic code in the starter, IE: spin one motor, just to know that it's working well before people start writing code.

jgkamat (approved):
I think the changes here are mostly good, although I really don't like vendoring libraries in binary form, it's probably going to break for a lot of people.

Later on, maybe we could put it into a submodule or have a script to install the library.

We'll see how this works today, I guess.

barulicm:
@jgkamat I totally agree about not shipping binaries in general. We're trying to keep it as brain dead simple as possible for the freshmen. I don't expect it to break for a lot of people. There may be a few, but the vast majority of them will be coming in the pretty default installs of their OS's.

hardware_applications/week1/starter_code.cpp
barulicm: These hardware demos are designed to exercise concepts we teach each week. So, this week, we want them to go through writing the main signature and just basic C++ setup. 
minimal "spin a motor" code is really straightforward and would be a good first step for guiding people through the demo:
```
int main() {
    RJRobot robot;
    robot.SetMotor(MotorPort::A, 200);
    return 0;
}
```

ebarnette (approved):

jgkamat:
> We're trying to keep it as brain dead simple as possible for the freshmen.

So don't ship binaries, they force you to manually re-compile the binary for most machines which is the opposite of brain-dead. Instead make a proper dependency, so we actually could acheive that goal.

PR #72, opened 2017-09-25
Week two code fixes

JasonGibson274:


code/week2/week2.cpp
```diff
@@ -1,6 +1,6 @@
 // make sure to compile with g++
 // to compile do
-//      g++ -o week2.out *.cpp
+//      g++ -o week2.out -std=c++14 *.cpp
```
jgkamat, line 3: Full support for c++14 was added in gcc5, which is pretty old at this point, but I'm not sure if everyone has it. Ubuntu 14.04 ships with gcc4. Does the windows and mac reccomended install come with a gcc new enough?

jgkamat: Notably, the weird windows linux binary thing comes with 14.04, as far as I know, so if we use c++14 features it might trip them up.

JasonGibson274: So we use c++ 14 material in the following weeks. This complies on my mac and a lot of people installed MinGW this week so I think this is a safe assumption.

jgkamat (approved):

PR #74, opened 2017-09-25
Add dependency on STSL for individual targets

jgkamat:
When compiling with multithreading, the build fails the first time, but succeeds on a second run (à la IGVC).

This patch fixes that by adding the STSL as a dependency of individual build targets.


barulicm:
Good catch

PR #82, opened 2017-09-30
Fix week3 git slides

joshhting:


slides/week3/git.org
```diff
@@ -25,12 +28,23 @@
...
-- git history is a *graph*
+** Git as a Graph
+- Git history is a *graph*
+- A graph is simply a collection of nodes connected by edges
 - Each commit is a node on the graph
-- Ways to see it as a graph, rather than a line of commits
-  - ~git~ porcelain
+- Subsequent commits are linked by edges
+
+** Git as a Graph cont.
```
JasonGibson274: Can we have a speaker note to mentioning that this is a installed package and is not installed by default.

slides/week3/git.org
```diff
@@ -25,12 +28,23 @@
...
+- Some useful tools to visualize your repo's git history as a graph:
+  - ~git porcelain~
     #+BEGIN_SRC sh
       git log --oneline --graph --all --decorate
```
JasonGibson274: might be nice to show this on the software training repo since ours branches a lot.

JasonGibson274 (changes requested):
Just a quick fix but looks good otherwise.

slides/week3/git.org
joshhting: Isn't that what the output on this slide is showing?

JasonGibson274: Valid point

JasonGibson274 (dismissed):

slides/week3/git.org
```diff
@@ -25,12 +28,23 @@
...
+- Git history is a *graph*
+- A graph is simply a collection of nodes connected by edges
 - Each commit is a node on the graph
-- Ways to see it as a graph, rather than a line of commits
-  - ~git~ porcelain
+- Subsequent commits are linked by edges
+
+** Git as a Graph cont.
+- Some useful tools to visualize your repo's git history as a graph:
+  - ~git porcelain~
```
jgkamat: x

joshhting: You mean change this line to just `- Git Porcelain`?

jgkamat: Yup, that worksf or me.

slides/week3/git.org
```diff
@@ -43,110 +60,84 @@
...
 - Create branches with
-
   #+BEGIN_SRC sh
     # git branch <BRANCH_NAME>
 
     # Create two branches, starting from the current commit
 
-    git branch stable
-    git branch beta
+    git branch branch1
```
jgkamat: Any reason why the branches were renamed? I was trying to make a realistic story and setting on why branches should be used (so there is a real world example right there)  rather than just using them. See the commands below.

slides/week3/git.org
```diff
@@ -43,110 +60,84 @@
...
-    git checkout beta
+    git checkout branch2
     echo "This feature is unstable!" > beta_release.txt
     git add -A
     git commit -m "Add beta_release.txt"
   #+END_SRC
-- Since the beta branch was *behind* the stable branch, this creates a 'fork' in the commit history
-- Notice, master has not moved at all (since no commits have been made on it)
-  #+BEGIN_SRC
-    1c65f7b * beta Add beta_release.txt
```
jgkamat, line 111: Any reason why the graphs were removed? Trying to hide git's graph structure leads to MUCH more confusion down the road (most people try to hide the graph structure and treat it as a line, which is not the case). I was confused for a good year on rebasing and merging due to this.

slides/week3/git.org
```diff
@@ -25,12 +28,26 @@
 - Branches are what people have the most trouble with
   - _*Ask questions /please/!*_ This is a difficult topic.
 
-** Git As a Graph
+** Remotes vs Branches vs Forks
+- A fork or clone does a full copy of the git repo
+- A remote is a pointer to another copy of this git repo
+- A repository is a collection of *branches*
```
jgkamat: This is misleading - it implies that branches are fundamental to the internal structure to git, while in reality they are much more just pointers to the underlying structure. 

For example, you can use git entirely without branches, if you wished, but the repository would still exist.

slides/week3/git.org
```diff
@@ -25,12 +28,26 @@
...
-- Ways to see it as a graph, rather than a line of commits
-  - ~git~ porcelain
+- Subsequent commits are linked by edges
+
+** Git as a Graph cont.
+#+BEGIN_NOTES
+These tools are not installed by default.
+#+END_NOTES
+- Some useful tools to visualize your repo's git history as a graph:
+  - ~git porcelain~
```
jgkamat: On second thought, this is kind of random, and I think we should just remove it (just say `git`).

jgkamat (changes requested):

slides/week3/git.org
JasonGibson274: they were renamed since they appeared to be keywords and we wanted to make sure they knew that a branch has an arbitrary name.

slides/week3/git.org
JasonGibson274: the text based renderings were confusing. I think adding speaker notes to pull up something like gitk would be better to show the structure instead of showing them text renderings that do not make any sense if you do not already know what a commit is (which they really do not).

slides/week3/git.org
jgkamat: Just name it to `stableRj` and `betaRj` then? Or similar? It's possible to get both.

jgkamat (approved):
Seems good to me

PR #98, opened 2017-10-23
Week Five Hardware

barulicm:
Given the general lessons we've been learning from doing this for a few weeks now, I tried to make this a simpler demo while still touching both classes and shared_ptrs. It feels a little bit on the too simple side of things, so I may have over-corrected.


barulicm:
Also, weekFive should be updated with master to simplify this PR.

JasonGibson274:
While this is on the simple side I think that this is good. Can we remove the final file from what is being merged into master? They never try to code it they just copy the solution and giving them a blank file tends to help with retention in my opinion. Also it is such a simple class I think any instructor can write it from memory.

hardware_applications/week5/final_main.cpp
```diff
@@ -0,0 +1,63 @@
+#include <STSL/RJRobot.h>
```
JasonGibson274, line 1: Can we remove this file?

JasonGibson274 (changes requested):
Remove file

hardware_applications/week5/final_main.cpp
barulicm: ... If we do... the instructors will not have the code they actually need to guide people through? Unless I'm misunderstanding your request.

JasonGibson274: you understand me correctly

barulicm:
Ok, so eventually, we want the final code in the repo for people coming back to it in the near and far future.
Long term, we either have to find a way to solve the "just run the final code" problem in the classroom or find a different place to hide the final code during the class while making it available to people later in the semester.
I don't know how we execute the second option when all of our resources are centralized in this repo

hardware_applications/week5/final_main.cpp
barulicm: See longer comment on the main thread, written after I saw your longer comment in the main conversation thread.

JasonGibson274:
Add it back in after all of the sessions have happened. I agree this is a larger problem. A solution that I have used for ROS training is committing it until I get the solution I like. Them remove it commit that and when it is time to get the solution revert that commit and push.

barulicm:
And, without any ill will towards our instructors, I'm not convinced everyone will want to be remembering even this simple demo when trying to teach it, given some of the confusion and feedback I got from the early week demos.

barulicm:
I'm a bigger fan of the proposal brought forth in the last meeting I was at, where it was suggested that the robots not even be passed out until the code walkthrough is done.

JasonGibson274 (approved):

JasonGibson274:
That works too.

PR #104, opened 2017-10-25
Implements TicTacToe demo for week6.

barulicm:
This PR replaces the Week 6 CV code with the TicTacToe code we'd talked about.

The files we're giving them won't compile because the Board class is missing. Instead, comments are left behind giving instructions for what needs to be written.

The final solution code that actually works is zipped up in the solution.zip folder.  In theory, this zip is password protected under 'robojackets', but I don't think it actually worked.


code/week6/CMakeLists.txt
```diff
@@ -1,11 +1,6 @@
...
-find_package(OpenCV REQUIRED)
 
 set(CMAKE_CXX_STANDARD 14)
 
-add_executable(main main.cpp)
-add_executable(findBlue findBlue.cpp)
-
-target_link_libraries(main ${OpenCV_LIBS})
-target_link_libraries(findBlue ${OpenCV_LIBS})
+add_executable(ttt main.cpp Board.cpp Board.h)
```
jgkamat: I don't think header files should be added to the executable targets.

code/week6/main.cpp
```diff
@@ -1,93 +1,104 @@
...
+        case Marker::Empty:
+          return ' ';
+        case Marker::X:
+          return 'X';
+        case Marker::O:
+          return 'O';
+      }
+  };
+  std::string layout = " | | \n-----\n | | \n-----\n | | ";
+  layout[0] = charForMarker(board.getMarker(0,0));
```
jgkamat: Hard coding these positions smells bad, why not use printf instead (so this can be readable)?

IMO, this should probably be properly printed out in a loop, this approach is really hacky.

code/week6/main.cpp
```diff
@@ -1,93 +1,104 @@
...
-    // and each channel is a uchar
-    int nCols = my_image.cols * channels;
+void generateComputerMove(const Board &board, unsigned int &r_out, unsigned int &c_out) {
+  static random_device rand_dev;
+  static default_random_engine engine(rand_dev());
+  static uniform_int_distribution<unsigned int> uniform_dist(0,2);
+  do {
+    r_out = uniform_dist(engine);
+    c_out = uniform_dist(engine);
+  } while(board.getMarker(r_out, c_out) != Marker::Empty);
```
jgkamat, line 15: Won't this infinite loop if the board is full?

code/week6/Board.cpp
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * isOver
+ * @return true if the game is over, false otherwise
+ * @note The game is over if no more moves can be made or if one player has won the game
+ */
+
+
+/**
+ * getWinner
+ * @return The current winner, or Empty if tie or in-progress game
```
jgkamat: Empty if tie.

jgkamat (changes requested):

code/week6/CMakeLists.txt
barulicm: Good catch. CLion auto adds them as part of the new file wizard.

code/week6/main.cpp
barulicm: The main method checks isOver before calling this method, so in practice board will not be full.
Were this a more generic code base, yes, we'd want to check that here.

code/week6/main.cpp
barulicm: That's fair. I was being lazy

JasonGibson274 (dismissed):
Overall this looks good to me. You used some things that they will not know (regex, enums) but I think that exposing them to new concepts is good since they will not understand their entire code base when they start working on teams anyway. Also the password protected zip file worked on my computer.

code/week6/main.cpp
```diff
@@ -1,93 +1,107 @@
...
+}
 
-    // iterate over an image the safe way
-    for(auto it = my_image.begin<cv::Vec3b>(); it != my_image.end<cv::Vec3b>(); it++) {
-        std::cout << "Blue = " << std::to_string((*it)[0]) << std::endl;
-        std::cout << "Green = " << std::to_string((*it)[1]) << std::endl;
-        std::cout << "Red = " << std::to_string((*it)[2]) << std::endl;
+void getMoveFromPlayer(unsigned int &r_out, unsigned int &c_out) {
+  static regex coordinate_patter("[0-3],[0-3]");
```
jgkamat: Shoudln't this be `[0-2]`, 3 is an invalid index if we are 0 indexed.

I also kind of dislike the part below it, (not verifying `find` is actually found, not catching `stoi` possible errors, but this regex is strict enough that it's probably fine).

jgkamat (dismissed):
Overall, I think this is good, just one small issue.

Do you think we should provide function stubs for some of the functions, to start off, or is that part of the excersize?

code/week6/main.cpp
barulicm: This definitely should be 0-2
And yes, the regex here catches most potential issues. It's not bullet proof, but it'll work.

barulicm:
@jgkamat I think writing the full class (stubs and all) is part of the exercise for this week.

jgkamat (approved):

joshhting (approved):
LGTM!

joshhting:
So the .zip isn't password protected, but the individual files within it are so it's all good

PR #110, opened 2018-08-24
Installs needed dependencies in circleci image before running tests

ebretl:
It would be better to get a new image with everything we need already installed, but this works a temporary fix at least

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,6 +7,7 @@ jobs:
       - image: silex/emacs:25.1
     steps:
       # Install cask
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -7,6 +7,7 @@ jobs:
       - image: silex/emacs:25.1
```
barulicm: Try changing this to silex/emacs:master-dev

Their dev variants come with cask and python installed already. That means we can probably also get rid of the cask installation line below.

barulicm (approved):

PR #134, opened 2018-09-24
Stable rj

reagankan:


ebretl:
Nice job. At this point someone would merge your PR into the rj/master branch. But since this is just an exercise, I'm going to close it instead. If you make a PR to the JR repository (github.com/robojackets/jr) you can get code merged in

PR #141, opened 2019-09-08
Create project 1, "drive_in_line.cpp"

chachmu:
This will need to be pointed to master once Barulic's branch merged. I'm not 100% sure exactly how STSL is going to be set up this year but I based this off of last years code so it should be close.

The description of the project mentions that they should change a number to see what happens, should we just let them pick any number or should I make a variable for them to change?

Also, my final set of questions, are we writing an explanation of the exact steps they should be following somewhere? If so, where is that?

### DIFF
--- a/projects/exercises/01/CMakeLists.txt
+++ b/projects/exercises/01/CMakeLists.txt
@@ -1,3 +1,0 @@
-add_executable(drive_in_line drive_in_line.cpp)
-add_dependencies(drive_in_line STSL)
-target_link_libraries(drive_in_line ${STSL_LIBS})
--- a/projects/exercises/01/CMakeLists.txt
+++ b/projects/exercises/01/CMakeLists.txt
@@ -1,0 +1,3 @@
+add_executable(drive_in_line drive_in_line.cpp)
+add_dependencies(drive_in_line STSL)
+target_link_libraries(drive_in_line ${STSL_LIBS})
--- a/projects/exercises/01/drive_in_line.cpp
+++ b/projects/exercises/01/drive_in_line.cpp
@@ -2,7 +2,7 @@
 
 int main()
 {
-    RJRobot robot(); //Initialize a RJ Robot
+    RJRobot robot; //Initialize a RJ Robot
     
     robot.setDriveMotors(1.0, 1.0); //Drive both robot motors forward at max speed
     robot.wait(2000ms); //Wait for 2000 milliseconds (2 seconds)
--- a/projects/exercises/01/drive_in_line.cpp
+++ b/projects/exercises/01/drive_in_line.cpp
@@ -4,10 +4,11 @@ using namespace std;
 
 int main()
 {
-    RJRobot robot(RobotType::REAL); //Initialize a RJ Robot that controls a real robot
+    RJRobot robot(); //Initialize a RJ Robot
     
-    robot.setDriveMotors(127, 127); //Drive both robot motors forward at max speed
-    robot.Wait(2000ms); //Wait for 2000 milliseconds (2 seconds)
-    robot.StopMotors(); //Stop the motors
+    robot.setDriveMotors(1.0, 1.0); //Drive both robot motors forward at max speed
+    robot.wait(2000ms); //Wait for 2000 milliseconds (2 seconds)
+    robot.stopMotors(); //Stop the motors
 
+    return 0;
 }
--- a/projects/exercises/01/drive_in_line.cpp
+++ b/projects/exercises/01/drive_in_line.cpp
@@ -1,0 +1,13 @@
+#include <STSL/RJRobot.h>
+
+using namespace std;
+
+int main()
+{
+    RJRobot robot(RobotType::REAL); //Initialize a RJ Robot that controls a real robot
+    
+    robot.setDriveMotors(127, 127); //Drive both robot motors forward at max speed
+    robot.Wait(2000ms); //Wait for 2000 milliseconds (2 seconds)
+    robot.StopMotors(); //Stop the motors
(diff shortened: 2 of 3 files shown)


projects/exercises/01/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_executable(drive_in_line drive_in_line.cpp)
```
barulicm: No CMake on the robots this year. The IDE already has a "Runner" script in it that pulls in all of the libraries needed.

projects/exercises/01/drive_in_line.cpp
```diff
@@ -0,0 +1,13 @@
+#include <STSL/RJRobot.h>
+
+using namespace std;
+
+int main()
+{
+    RJRobot robot(RobotType::REAL); //Initialize a RJ Robot that controls a real robot
+    
+    robot.setDriveMotors(127, 127); //Drive both robot motors forward at max speed
+    robot.Wait(2000ms); //Wait for 2000 milliseconds (2 seconds)
```
barulicm: Please refer to [RJRobot.h](https://github.com/RoboJackets/stsl/blob/master/include/STSL/RJRobot.h) for the current API.
Wait is not in the current API, but I can add it if we want. It was in last year's to be a cross-platform wait command. We technically don't need that this year since we're back to single platform. Of course, if we just like that syntax over, say, usleep(), I can add a simple wait() to RJRobot.

projects/exercises/01/drive_in_line.cpp
```diff
@@ -0,0 +1,13 @@
...
+    robot.StopMotors(); //Stop the motors
```
barulicm:
```suggestion
    robot.stopMotors(); //Stop the motors
```

projects/exercises/01/drive_in_line.cpp
```diff
@@ -0,0 +1,13 @@
...
+
+}
```
barulicm, line 12:
```suggestion
    return 0;
}
```

projects/exercises/01/drive_in_line.cpp
```diff
@@ -0,0 +1,13 @@
+#include <STSL/RJRobot.h>
+
+using namespace std;
+
+int main()
+{
+    RJRobot robot(RobotType::REAL); //Initialize a RJ Robot that controls a real robot
```
barulicm:
```suggestion
    RJRobot robot; //Initialize a RJ Robot
```

No sim support in STSL this year.

projects/exercises/01/drive_in_line.cpp
```diff
@@ -0,0 +1,13 @@
...
+    
+    robot.setDriveMotors(127, 127); //Drive both robot motors forward at max speed
```
barulicm:
```suggestion
    robot.setDriveMotors(1.0, 1.0); //Drive both robot motors forward at max speed
```

Drive motor valid ranges are -1.0 to 1.0

barulicm (changes requested):

barulicm:
* You can just let them change any number. In the code you've written so far, we've got two motor speeds and a wait duration to play with.
* Yes, you'll need to write out instructions, but not for the students. We need to give the instructors a document that lays out how they should present the exercise and any troubleshooting tips that are specific to each project. I'm not sure where those instructions should live. Maybe a sibling directory to `projects/exercises`, (maybe `projects/instructions`) with markdown files?

Zolkin1:
> * You can just let them change any number. In the code you've written so far, we've got two motor speeds and a wait duration to play with.
> * Yes, you'll need to write out instructions, but not for the students. We need to give the instructors a document that lays out how they should present the exercise and any troubleshooting tips that are specific to each project. I'm not sure where those instructions should live. Maybe a sibling directory to `projects/exercises`, (maybe `projects/instructions`) with markdown files?

I think that a sibling directory is probably the way to go. I think we should be able to talk about the exercises together, so I'm not convinced we need a "how to present it" section. But a troubleshooting guide would be good.

barulicm:
Even if all of our trainers this year get to talk about it, providing clear presentation instructions makes it easier for us to reuse these projects in the future.

Zolkin1:
That's fair, so I am thinking that we can write something now, but know that in the future (spring semester) we can go back and re-write any necessary parts after we have the experience of teaching it. I'm just thinking that this is going to be a big learning year for us with new robots and videos. So hopefully we can keep track of everything to be better next year.

projects/exercises/01/drive_in_line.cpp
```diff
@@ -0,0 +1,14 @@
+#include <STSL/RJRobot.h>
+
+using namespace std;
```
barulicm: I don't think we need this in this project.

Zolkin1: @chachmu can you remove this, then I'll merge it?

projects/exercises/01/drive_in_line.cpp
```diff
@@ -0,0 +1,12 @@
+#include <STSL/RJRobot.h>
+
+int main()
+{
+    RJRobot robot(); //Initialize a RJ Robot
```
barulicm:
```suggestion
    RJRobot robot; //Initialize a RJ Robot
```

Last change I think we need before merging, but this line actually invokes the [most vexing parse](https://en.wikipedia.org/wiki/Most_vexing_parse). The suggested change should fix that.

Zolkin1 (approved):
looks good. I checked yesterday that this compiles/runs on the robots

barulicm (approved):

PR #142, opened 2019-09-15
Adds pthread to the RoboJackets C++ runner.

barulicm:
STSL now uses the standard threading library, which means we need to link against pthread in our Cloud9 runner. This PR adds pthread to the runner.


oswinso (approved):
lgtm

Zolkin1 (approved):
Looks good

PR #146, opened 2020-08-23
setup cmake configuration within the assignments section

JasonGibson274:
I'm bad a cmake, so please let me know if I did not do something right. I think eventually we will want to make include and src directories in each of the weeks, but that will only be once they are using multiple files.

The goal I was going for was that all you needed to do was copy the CMakeLists.txt that I have setup in the example places into whatever folder you need without having to change the CMakeLists.txt itself. So for exercise 2.1 you would just copy the CMakeLists.txt from exercises 1.1 or 1.2 or the project and it will handle the rest from there.


barulicm:
So, honestly, with an eye towards these exercises eventually being split up into individual repos for github classroom (even if its not this year), I'd be ok with each week or even each exercise just being its own cmake project. Do we feel like its a big advantage to have a central build point for all of the exercises?

CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+cmake_minimum_required(VERSION 3.8)
+project(MPPIGeneric LANGUAGES C CXX)
+
+set(CMAKE_CXX_STANDARD 11)
```
barulicm:
```suggestion
set(CMAKE_CXX_STANDARD 17)
```

.gitignore
```diff
@@ -8,3 +8,7 @@ CMakeFiles/
 projects/.c9/*
 *.o
 
+*.idea
+cmake-build-debug
```
barulicm, line 12:
```suggestion
cmake-build-*
```

So we can ignore all built types

JasonGibson274:
The only advantage is making the build process match catkin better (everything is compiled in one place) and that's tenuous at best. I'm fine with splitting all of them into separate projects. We can make that call at the meeting.

kylestach:
@JasonGibson274 when is this expected to be ready to merge?

PR #148, opened 2020-09-02
Project 2.4

barulicm:
This adds the files for project 2 (traffic light state machine).

This includes:

- Instructions
- Starter code
- Solution code
- Test script
- Test case data


assignments/week_2/project_2/main.cpp
```diff
@@ -0,0 +1,23 @@
...
+{
+    // TODO Initialize your state machine class here
+    while(true)
+    {
+        std::string input;
+        std::getline(std::cin, input);
+        if(input.size() != 2)
+            break;
+
+        // TODO Parse the input digits
```
kylestach: Are we having them do the parsing themselves? I'm not opposed to that, but I think it would be good to provide them with a little bit more structure here.

assignments/week_2/project_2/readme.md
```diff
@@ -0,0 +1,135 @@
+# Project 2: Traffic Light State Machine
+
+In this project, you'll be creating software to control the traffic lights for a road intersection. State machines are
+actually super useful for traffic lights! Instead of having to manage all of the possible conditions that might control
```
kylestach: Maybe mention that robots use state machines too, and give an example?

assignments/week_2/project_2/main.cpp
kylestach: Maybe we could give them a `struct` declaration that contains the six booleans?

assignments/week_2/project_2/readme.md
```diff
@@ -0,0 +1,135 @@
...
+
+Our controller always allows cars going north or south to proceed if they are waiting. If our expectation that traffic
+going north-south is uncommon turns out to be incorrect, what problem might this behavior cause? How would you change
+the state machine for the controller to address this problem?
+
+Do you feel that structuring this code using a class helped you understand the code better? Why or why not?
+
+Did you use access specifiers (public, private, etc.) to protect your class's internal data? If not, how do you make
+sure that your state variable only changes in accordance with your transition rules? Would you be certain of that
+guarantee if you handed your class over to someone else to use (and potentially abuse)?
```
kylestach: I like these reflection questions

assignments/week_2/project_2/readme.md
```diff
@@ -0,0 +1,135 @@
...
+
+We've included an example solution state machine here, but make sure you take time to try to work through this part
+yourself. This is here to help you along if you really, truly get stuck.
+
+<details>
+<summary>Solution</summary>
+
+![Example state machine](state_machine.png)
+
+</details>
```
kylestach: Maybe have two levels of hint, where the first one tells you what the states should be and the second one is the full state machine?

kylestach (approved):
Minor comments, looks really good!

PR #149, opened 2020-09-02
Exercise 1.2

Daniel-Martin576:
Files for exercise 1.2
Apologize for the delay

## What we want to test:
- Knowledge of basic C++ syntax
- Variable declaration 
- How and when to use loops
- How to index into and iterate through arrays
- How to use conditionals
## Exercise 1.2: Array Reversal and Palindrome Check
- Reverse an int array
- Check if a char array is a palindrome and print the result 

The original tasking was "reversing an array" but I thought that was too little for the exercise so I added the palindrome check so they could have extra practice. Let me know if you want me to change the array to vector (at that point they can use `std::reverse`) or if you want me to add more. I also wasn't sure if I should have added the CMakeList.txt since I didn't remember if we agreed to do g++ or CMake. 

Also, I think that Exercise 1.2 is a decent bit easier than Exercise 1.1 so maybe we could switch them in order.


assignments/week_1/exercise_1_2/exercise_1_2.cpp
```diff
@@ -0,0 +1,34 @@
+// Iterative C++ program to reverse an array and check if string is palindrome
+#include <stdio.h>
+#include <string.h>
```
barulicm: Why do we need these two C headers for this exercise?

assignments/week_1/exercise_1_2/exercise_1_2.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include <iostream>
+
+int main()
+{
+    // Part 1: Array Reversal
+    int arr[] = {1, 2, 3, 4, 5, 6};
```
barulicm: If we don't want to teach `std::array`, we should stick to `std::vector`.
Codecademy doesn't actually cover language-level arrays... We should discuss as a group for real which containers we want to teach.

assignments/week_1/exercise_1_2/exercise_1_2.cpp
```diff
@@ -0,0 +1,34 @@
...
+    int n = sizeof(arr) / sizeof(arr[0]);
```
barulicm: In general, we should be lead by example with descriptive variable names. That said, in this case, I think we can just replace `n` with `arr.size()` when it's a vector.

assignments/week_1/exercise_1_2/exercise_1_2.cpp
```diff
@@ -0,0 +1,34 @@
...
+    // WRITE YOUR STRING REVERSAL HERE
+    
+    // Print the Reversed array
+    std::cout << "Reversed Array: ";
+    for (int i = 0; i < n; i++)
+        std::cout << arr[i] << " ";
+    std::cout << std::endl;
+
+    // Part 2: Palindrome Check
+    char str[] = "abbcdcbba";
```
barulicm: I think we should prefer `std::string` over character arrays in our exercises.

assignments/week_1/exercise_1_2/exercise_1_2.cpp
```diff
@@ -0,0 +1,34 @@
...
+int main()
+{
+    // Part 1: Array Reversal
+    int arr[] = {1, 2, 3, 4, 5, 6};
+    int n = sizeof(arr) / sizeof(arr[0]);
+
+    // Print original array
+    std::cout << "Original Array: ";
+    for (int i = 0; i < n; i++)
+        std::cout << arr[i] << " ";
```
barulicm: I think these should be range-based for loops, since we're iterating over the entire array. This would also eliminate the 
 one-letter variables.

```suggestion
    for (int elem : arr)
        std::cout << elem << ' ';
```

barulicm (changes requested):
This looks good. I think the code should lean more into C++. There are a couple heavy C idioms in the current code.

In line with some of the discussions we've had at the meetings, and with an eye towards these eventually being autograded, I'd recommend splitting the two parts (reversing and palindromes) into separate executables. They should take in an array of numbers / string via `std::cin` (or from command line args) and print their results.

I'd also really like us to add notes or review sections to these early exercises that show what this code would look like using the standard library. This exercise in particular has a lot of opportunities for that. 
- `std::swap`
- `std::reverse`

You can even do the whole palindrome check in one line.
```
std::equal(str.begin(), str.begin() + str.size()/2, str.rbegin());
```
(I'm not saying you have to "teach" or explain all of these. We should let students know that the language and library provide more powerful tools for common tasks like these, and link them to the relevant cppreference pages for those functions.)

assignments/week_1/exercise_1_2/exercise_1_2.cpp
Daniel-Martin576: Removed

Daniel-Martin576:
I tried to address all the issues you mentioned. Let me know if you would like me to change anything else.

assignments/week_1/exercise_1_2/exercise_1_2.md
```diff
@@ -0,0 +1,231 @@
+# Exercise 1.2
+Welcome to your second exercise of RoboJackets software training! This exercise will be testing your knowledge of basic C++ syntax and vector manipulation. For this exercise we expect you to have watched or know the content of the Codecademy C++ course up through vectors.
+
+# Exercise Objective
+In this exercise, you will be reversing an vector and checking if a string is a palindrome. The purpose of this exercise is to get you familar with working with vectors and strings. This exercise should help solidify your knowledge of loops, variables, conditionals, and vectors. _We aren't going to be using any new C++ standard library functions in the instructions, but feel free to use them if you want._
+
+# Starting Code
+We have already given you a main() function to write your code in and have included the proper libraries for you to use. We have also set up most of variable initilization and commented where you will write your code.
+
```
barulicm, line 9: Give at least one example of what the input should look like and what the expected output for that input would be.
Make sure this example makes it clear what "end list with non-integer" means.

assignments/week_1/exercise_1_2/exercise_1_2.md
```diff
@@ -0,0 +1,231 @@
...
+#include <iostream>
+#include <vector>
+
+int main()
+{
+    // Part 1: Vector Reversal
+
+    // Input vector from command-line
+    std::vector<int> list;
+    int input;
```
barulicm: We should initialize all variables in our examples to encourage that habit in the new members.
```suggestion
    int input = 0;
```

assignments/week_1/exercise_1_2/exercise_1_2.md
```diff
@@ -0,0 +1,231 @@
...
+```c++
+#include <iostream>
+
+int main()
+{
+    // Part 2: Palindrome Check
+
+    // Input string from command-line
+    std::string str;
+    std::cout << "Ender String:" << std::endl;
```
barulicm: Typo
```suggestion
    std::cout << "Enter String:" << std::endl;
```

assignments/week_1/exercise_1_2/exercise_1_2.md
```diff
@@ -0,0 +1,231 @@
...
+        if (str[left] != str[right])
+        {
+            // MORE CODE
+        }
+    }
+  ```
+
+</details>
+
+5. If the elements are different, then the string should be marked are not a palindrome and you should leave the loop
```
barulicm:
```suggestion
5. If the characters are different, you should set `isPalindrome` to false and leave the loop.
```

assignments/week_1/exercise_1_2/exercise_1_2.md
```diff
@@ -0,0 +1,231 @@
...
+        start++;
+        end--;
+    }
+  ```
+
+</details>
+
+Now try to run the program and test that the vector gets reversed!
+
+
```
barulicm, line 120: Add an introduction for this part of the exercise. Give a brief explanation of what a palindrome is and example input/output for this part.

assignments/week_1/exercise_1_2/exercise_1_2.md
```diff
@@ -0,0 +1,231 @@
...
+</details>
+
+Now run the program and test multiple strings to check your program works!
+
+# Review
+Great Job! You successfully manipulated a vector and string in this exercise. Although, these intructions explains one method to do vector reversal and palindrome
+checking, there exists much easier ways to do these operations using C++ standard library functions. For example, `std::reverse(list.begin(), list.end());` would
+reversal the vector in just one line without any hassle! Also, `str == string(str.rbegin(), str.rend())` would check if `str` is a palindrome in one line as well.
+The standard library is powerful tool for writing C++ code and as you progress through future weeks' exercises you become more and more familiar with it!
```
barulicm: Suggestion: (can't use the normal suggestion feature because my suggested text includes code blocks.

Great Job! You successfully manipulated a vector and string in this exercise. Although, these instructions explain one method to do vector reversal and palindrome checking, there are much easier ways to do these operations using C++ standard library functions. 

For example, [`std::reverse()`](https://en.cppreference.com/w/cpp/algorithm/reverse) lets you easily reverse any standard container.

```
std::reverse(list.begin(), list.end());
```

For palindrome checking, there are multiple one-line solutions using the standard library features. Standard containers (including `std::string`) have reverse iterators that let you iterate over the container in reverse order. So the palindrome check could be written like this:

```
bool is_palindrome = ( str == string(str.rbegin(), str.rend()) );
```

Or even in a way that's generic for any standard container:

```
bool is_palindrome = std::equal(str.begin(), str.begin() + str.size()/2, str.rbegin());
```

Checkout [`std::string::rbegin()`](https://en.cppreference.com/w/cpp/string/basic_string/rbegin) and [`std::string::rend()`](https://en.cppreference.com/w/cpp/string/basic_string/rend) for documentation.

The standard library is a powerful tool for writing C++ code. As you progress through future weeks' exercises, you'll become more and more familiar with it!

assignments/week_1/exercise_1_2/exercise_1_2.md
```diff
@@ -0,0 +1,231 @@
...
+  ```
+
+</details>
+
+2. Create a variable (`isPalindrome`) that will store the result of the palindrome check
+<details>
+  <summary>Answer</summary>
+
+  ```c++
+    bool isPalindrome = true;
```
barulicm: Per our FAQ, our code should follow the ROS code style guides.
```suggestion
    bool is_palindrome = true;
```

barulicm (changes requested):

barulicm (approved):
Thanks for making all of those changes!

kylestach:
My only comment before approval would be: please use braces for single line conditionals/loops! I don't want them to think that this is Python/braces are optional.

Daniel-Martin576:
I definitely like
```c++
for (int v : list) 
    std::cout << v << " ";
```
more than 
```c++
for (int v : list) 
{
    std::cout << v << " ";
}
```
but sure, I can change it

Daniel-Martin576:
Pending @kylestach approve for merging

PR #151, opened 2020-09-06
Exercise 2.1

JasonGibson274:
This contains exercise 2.1 which covers converting project 1.3 into a method. They first do this in the same file, then they do it in a separate .cpp and .h file.

This depends on project 1 so that is why I am merging into that branch. The solution code will be updated if people disagree with how it was done.


assignments/week_2/exercise_2_1/exercise_2_1.cpp
```diff
@@ -0,0 +1,5 @@
+#include <vector>
+
+int main() {
+
```
kylestach, line 4: Maybe leave the `readInVector` stuff? Or do we just assume they'll copy their entire solution in here? If it's the latter I'd leave a comment to mention that.

assignments/week_2/exercise_2_1/exercise_2_1.md
```diff
@@ -0,0 +1,173 @@
...
+std::vector<double> readInVector(std::string s) {
+  int prev_location = 0;
+  int next_location = 0;
+  std::vector<double> result;
+  while(s.find(',', prev_location) != std::string::npos) {
+    next_location = s.find(',', prev_location);
+    //std::cout << "prev_location: " << prev_location << std::endl;
+    //std::cout << "next_location: " << next_location << std::endl;
+    // substr [,]
+    //std::cout << s.substr(prev_location, next_location - prev_location) << std::endl;
```
kylestach: Commented out code?

assignments/week_2/exercise_2_1/exercise_2_1.md
```diff
@@ -0,0 +1,173 @@
...
+  for(int i = 1; i < x.size(); i++) {
+    std::cout << ", " << x[i];
+  }
+  std::cout << "}" << std::endl;
+
+  std::cout << "w: {" << w[0];
+  for(int i = 1; i < w.size(); i++) {
+    std::cout << ", " << w[i];
+  }
+  std::cout << "}" << std::endl;
```
kylestach, line 57: Could we make them write two functions: one to do the convolution and one to print out a vector?

assignments/week_2/exercise_2_1/exercise_2_1.cpp
JasonGibson274: Leaving this open to make sure the change is sufficient.

kylestach (approved):

PR #153, opened 2020-09-10
Exercise2 2

JasonGibson274:
This contains exercise 2.2 which is implementing some simple methods in a class that is already written.

I think overall this is okay, I want someone to double check I am not pushing things too hard with the code. I took it from code Matt wrote 2 or so years ago for week 7, so it uses things we have definitely not covered, but I think with some extra explanations it should be fine.

This should be more of a sanity check than a full review right now. I want to expand the explanations, but the structure should remain about the same.


JasonGibson274:
This should be more of a sanity check than a full review right now. I want to expand the explanations, but the structure should remain about the same.

assignments/week_2/exercise_2_2/exercise_2_2.md
```diff
@@ -0,0 +1,347 @@
...
+enum class Marker {
+  Empty,
+  X,
+  O
+};
+```
+
+This is a definition of an enum, think of it like a special variable type with
+only the listed values as possibilities. This allows you to convey semantic meaning
+in code.
```
kylestach:
```suggestion
This is a definition of an enum, think of it like a special variable type that can only hold
one of the listed options as a value.
```

kylestach (commented):
Cursory glance says this is mostly okay, but please merge master into this so I can see only the new diff.

assignments/week_2/exercise_2_2/Board.cpp
```diff
@@ -0,0 +1,117 @@
...
+ *       winner=Empty
+ *       for i=0 to 2
+ *         if cell[x,i]=player then col++
+ *         if cell[i,y]=player then row++
+ *         if cell[i,i]=player then diag++
+ *         if cell[i,2-i]=player then rdiag++
+ *       if row=n or col=n or diag=n or rdiag=n then winner=player
+ */
+void Board::updateWinner(int last_r, int last_c) {
+  int hor = 0;
```
kylestach, line 98: They should implement this, right?

kylestach: NVM. Just realized that they're only implementing a few methods.

kylestach (approved):

PR #154, opened 2020-09-10
Exercise 3.2

HussainGynai:
Linked List exercise (TODO: add answers using spoiler tags under each step)

### DIFF
--- a/assignments/week_3/exercise_3_2/exercise_3_2.hpp
+++ b/assignments/week_3/exercise_3_2/exercise_3_2.hpp
@@ -1,0 +1,31 @@
+#include <string>
+#include <vector>
+
+class Node {
+public:
+	std::string name;
+	Node *next;
+};
+
+class LinkedList {
+private:
+	Node *head;
+public:
+
+	LinkedList();
+
+	void add_to_front(std::string name);
+
+	void add_to_back(std::string name);
+
+	void add_at_index(std::string name, int index);
+
+	void remove_from_front();
+
+	void remove_from_back();
+
+	void remove_at_index(int index);
+
+	void print_names();
+
+};
--- a/assignments/week_3/exercise_3_2/exercise_3_2.hpp
+++ b/assignments/week_3/exercise_3_2/exercise_3_2.hpp
@@ -1,15 +1,16 @@
 #include <string>
 #include <vector>
+#include <memory>
 
 class Node {
 public:
     std::string name;
-    Node *next;
+    std::unique_ptr<Node> next;
 };
 
 class LinkedList {
 private:
-    Node *head;
+    std::unique_ptr<Node> head;
 public:
 
     LinkedList();
--- a/assignments/week_3/exercise_3_2/exercise_3_2.hpp
+++ b/assignments/week_3/exercise_3_2/exercise_3_2.hpp
@@ -3,29 +3,29 @@
 
 class Node {
 public:
-	std::string name;
-	Node *next;
+    std::string name;
+    Node *next;
 };
 
 class LinkedList {
 private:
-	Node *head;
+    Node *head;
 public:
 
-	LinkedList();
+    LinkedList();
 
-	void add_to_front(std::string name);
+    void add_to_front(std::string name);
 
-	void add_to_back(std::string name);
+    void add_to_back(std::string name);
 
-	void add_at_index(std::string name, int index);
+    void add_at_index(std::string name, int index);
 
-	void remove_from_front();
+    void remove_from_front();
 
-	void remove_from_back();
+    void remove_from_back();
 
-	void remove_at_index(int index);
+    void remove_at_index(int index);
 
-	void print_names();
+    void print_names();
 
 };
--- a/assignments/week_3/exercise_3_2/exercise_3_2.hpp
+++ b/assignments/week_3/exercise_3_2/exercise_3_2.hpp
@@ -1,5 +1,4 @@
 #include <string>
-#include <vector>
 #include <memory>
 
 class Node {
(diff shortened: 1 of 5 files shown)


assignments/week_3/exercise_3_2/exercise_3_2_answer.cpp
```diff
@@ -0,0 +1,93 @@
...
+        }
+        curr->next = new_node;
+    } else {
+        head = new_node;
+    }
+    return;
+}
+
+void LinkedList::add_at_index(std::string name, int index) {
+    Node *new_node = new Node();
```
kylestach: We don't want to teach them to use `new` like this. I think probably the sanest option is to just make each `next` pointer a `unique_ptr`, because it's a singly-linked-list.

kylestach (changes requested):

assignments/week_3/exercise_3_2/exercise_3_2.hpp
```diff
@@ -0,0 +1,31 @@
+#include <string>
```
JasonGibson274, line 1: Did we cover hpp? also I don't see any implementation in here so why not just .h?

assignments/week_3/exercise_3_2/exercise_3_2.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <vector>
+
+class Node {
+public:
+    std::string name;
+    Node *next;
```
JasonGibson274: Are we Node* or Node * people?

assignments/week_3/exercise_3_2/exercise_3_2.md
```diff
@@ -0,0 +1,111 @@
+# Exercise 3.2
+
+Welcome to exercise 3.2. This exercise will test your ability to create and use pointers. This exercise will expect you to know all the information in the Codecademy C++ course through the References and Pointers section. Our own video on Resources, Lifetime, and Ownership is not needed but will help you get a better understanding of pointers and their use.
+
+# Exercise Statement
+
+In this exercise you will finish writing the code needed to create a linked list. A linked list is a data structure that holds data, similar to a vector. A linked list is made up of nodes in a line each which contains two pieces of information:
+1. Whatever data you want to store in the linked list (in this case a string)
+2. A pointer to the next data
```
JasonGibson274: Next node?

assignments/week_3/exercise_3_2/exercise_3_2_answer.cpp
JasonGibson274: If code academy teaches new then I think this is fine, but we have to have the second half where we explain why not to do this. I don't like just yeeting in unique_ptr because it is the correct thing to do as an experienced C++ programmer, they aren't there yet. I am okay letting smaller things like this slide earlier on to reduce complexity as long as we follow up with best practice later on.

assignments/week_3/exercise_3_2/exercise_3_2.md
```diff
@@ -0,0 +1,111 @@
...
+
+    void print_names();
+
+};
+```
+
+# Steps
+### add_to_front:
+1. Create a new variable that is a pointer to a Node
+2. Set the name of the name of the new node to the name parameter passed into the function
```
JasonGibson274: set the name of the new node to the name parameter passed into the function by changing the string variable called name to the passed in value?

assignments/week_3/exercise_3_2/exercise_3_2.md
```diff
@@ -0,0 +1,111 @@
...
+3. Set the new node’s next variable to the head pointer of the linked list
+4. Set the head of the linked list to the new node pointer
+### add_to_back:
+1. Create a new variable that is a pointer to a Node
```
JasonGibson274: I think we need to call out the point that they need to allocate memory here, but IDK how to do that. because new variable to me reads like

Node node;

not 

Node* node = new Node();

assignments/week_3/exercise_3_2/exercise_3_2.md
```diff
@@ -0,0 +1,111 @@
...
+### add_at_index:
+1. Create a new variable that is a pointer to a Node
+2. Set the name of new node pointer to be the name parameter
+3. Create an if else statement that checks if the index is not 0
+4. In the if statement create a variable that is a pointer to a Node and define set it to the head pointer called curr
+5. In the if statement create a for loop that iterates index-1 number of times that sets curr to its next pointer
+6. Outside of the loop in the if statement set the new node’s next pointer to curr’s next pointer
+7. In the if statement set curr’s next pointer to the new node pointer
+8. In the else statement use the same technique in add to front to make the new node pointer the head
+9. remove_from_front:
```
JasonGibson274: This should be a heading right?

assignments/week_3/exercise_3_2/exercise_3_2.hpp
HussainGynai: In Codecademy they only ever use hpp

kylestach (approved):

assignments/week_3/exercise_3_2/exercise_3_2.md
```diff
@@ -0,0 +1,151 @@
...
+7. In the else statement set head to be the new node pointer again by using the std::move() function
+### add_at_index:
+1. Create a new variable that is a unique pointer to a Node
+2. Set the name of new node pointer to be the name parameter
+3. Create an if else statement that checks if the index is not 0
+4. In the if statement create a variable that is a pointer to a Node and define set it to the head pointer called curr using .get()
+5. In the if statement create a for loop that iterates index-1 number of times that sets curr to its next pointer using .get()
+6. Outside of the loop in the if statement set the new node’s next pointer to curr’s next pointer using std::move()
+7. In the if statement set curr’s next pointer to the new node pointer using std::move()
+8. In the else statement use the same technique in add to front to make the new node pointer the head
```
kylestach, line 117: I don't think you need to spell it out quite this much. Just link them to this page: https://visualgo.net/en/list

PR #160, opened 2020-09-20
Exercise2 3

JasonGibson274:
In this exercise they write a class that converts between different number representations (binary, hex, decimal). They are writing the class from scratch. We might want to call this a project since the code got more involved than I was expecting.


assignments/week_2/exercise_2_3/NumberConverter.cpp
```diff
@@ -0,0 +1,160 @@
+#include "NumberConverter.h"
+
+
+NumberConvereter::NumberConvereter(std::vector<int> vec) {
```
kylestach: Give `vec` a more descriptive name

assignments/week_2/exercise_2_3/exercise_2_3.md
```diff
@@ -0,0 +1,360 @@
...
+0x19 = 9 * 16^0 + 1*16^1 = 9 + 16 = 25
+```
+
+# Implementation
+
+Today we will be writing a class from scratch in order to convert between the
+different formats of the numbers. Furthermore, we will be keeping track of
+how many conversions we do between each type inside our class.
+
+## Starter Code
```
kylestach: Move this after "Creating a class" (or as a subsection)

assignments/week_2/exercise_2_3/exercise_2_3.md
```diff
@@ -0,0 +1,360 @@
...
+## Part 0 Creating a class
+
+create a .h and .cpp called whatever you would like, then you will need to add the .cpp
+to the cmake file like so (ours is named NumberConverter.cpp)
+
+```cmake
+add_executable(exercise_2_3 exercise_2_3.cpp NumberConverter.cpp)
+```
+
+to build and run your code call this in the directory with your code
```
kylestach: Remind them that any time they change the `CMakeLists.txt` they'll want to rerun `cmake ..`, but otherwise they can just run `make`

assignments/week_2/exercise_2_3/exercise_2_3.md
```diff
@@ -0,0 +1,360 @@
...
+Converting from binary and hex to decimal is straightforward. Implement two methods
+that take in strings and output an integer that is the decimal representation of
+the input. Remember you should be using the hexCharToNumber helper function to
+convert each digit. Think about how we were writing out the numbers earlier to help
+you along in your implementation.
+
+### Tips
+
+There is a function implemented called [std::pow()](https://en.cppreference.com/w/cpp/numeric/math/pow).
+This will make it easier to arbitrary powers.
```
kylestach: You accidentally a verb.

assignments/week_2/exercise_2_3/exercise_2_3.md
```diff
@@ -0,0 +1,360 @@
...
+```
+So if we combine out mod operators we end up with 7A2, now the C++ algorithms header
+gives us a function to reverse a string that can be called like this (make sure to have the header file.
+
+```c++
+#include <algorithm>
+// lotta cool code
+std::string str = "str";
+std::reverse(str.begin(), str.end());
+// str = "rts" now
```
kylestach, line 190: ??? why ???

assignments/week_2/exercise_2_3/exercise_2_3.md
```diff
@@ -0,0 +1,360 @@
...
+input: 2A7
+output type: d
+679
+decimal_inputs: 0
+binary_inputs: 0
+hex_inputs: 1
+decimal_outputs: 1
+binary_outputs: 0
+hex_outputs: 0
+```
```
kylestach, line 265: This isn't really the output they'll see though? They won't see the input from `example4_1.txt`.

kylestach: Also, 4_1???

assignments/week_2/exercise_2_3/NumberConverter.cpp
```diff
@@ -0,0 +1,160 @@
...
+  num_decimal_inputs_ = vec[0];
+  num_binary_inputs_ = vec[1];
+  num_hex_inputs_ = vec[2];
+  num_decimal_outputs_ = vec[3];
+  num_binary_outputs_ = vec[4];
+  num_hex_outputs_ = vec[5];
+}
+
+NumberConvereter::~NumberConvereter() {
```
kylestach: Typo: ~`Convereter`~ -> `Converter`

assignments/week_2/exercise_2_3/exercise_2_3.md
```diff
@@ -0,0 +1,360 @@
...
+Don't forget to include your class in the exercise_2_3.cpp.
+
+## Part 2 To Hex
+Now we will be writing two methods that convert from binary and decimal to hex.
+binary to decimal should take in a string and return a string, decimal to hex should
+take in an int and return a string.
+
+### Decimal To Hex Tips
+
+There is an elegant exploit to convert from decimal to hex. Let us say we have the number
```
kylestach: This isn't really an elegant exploit, is it?

Also, shouldn't these methods be written more like `parse{Hex|Dec|Bin}` -> int -> `display{Hex|Dec|Bin}`?

kylestach (changes requested):

assignments/week_2/exercise_2_3/exercise_2_3.md
JasonGibson274: why what?

assignments/week_2/exercise_2_3/exercise_2_3.md
JasonGibson274: part 4 example 1. I mean this is the output they should make it do since that is what we will check in github classroom?

assignments/week_2/exercise_2_3/exercise_2_3.md
JasonGibson274: removed the commentary of elegant, but I am not sure what you mean by the second comment?

kylestach: The second comment is:

It makes sense to implement `n` parsing functions and `n` display functions rather than creating `n(n-1)/2` end-to-end functions.

kylestach (approved):

PR #161, opened 2020-09-20
completed exercise 3.1

JasonGibson274:
This has a simple write up and a csv file. The goal is for them to get familiar with pointers by getting used to what it means to point at another variable. I made the worksheet short, but hopefully they will continue to play with more code until they get familiar.


kylestach (approved):

kylestach:
Can we import this into a google sheet for them to copy?

JasonGibson274:
Why a google sheet instead of locally?

It already exists btw https://docs.google.com/spreadsheets/d/1C5hVwX53Cib2dgHoiCmEBSTgXavHvFxQA9_DuuyzF6Y/edit?usp=sharing

PR #167, opened 2020-09-23
Project 3.5

Daniel-Martin576:
Project 3.5 (A* implemented in the style of project 3)
Pretty similar to project 3, so any comments on project 3 I will also change on this one (so they don't need to be duplicated). 
Similar to Project 3 PR, let me know what part of the implementation you would like me to change. Also, I feel the README is kinda lacking, so I would appreciate any suggestions.


assignments/week_3/project_3_5/graph_search_solution.hpp
```diff
@@ -0,0 +1,194 @@
...
+                {
+                    parent = cameFrom[parent];
+                    path.insert(path.begin(), parent);
+                }
+                if (verbose)
+                {
+                    for (Node *n : path)
+                        grid_[n->r][n->c] = "+";
+                    displayGrid();
+                }
```
JasonGibson274: I'd make this a method since it would match the pseudo code

JasonGibson274 (approved):
LGTM, small points that I don't think really matter.

PR #186, opened 2021-07-05
Coordinate Frame Student Code

JasonGibson274:
- This node transforms the aruco marker that is in camera frame into body frame (base_footprint).
- The transform combines two different transforms base_footprint -> camera_link and then camera_link -> camera_link_optical.
  - The first transform comes from tf
  - The second is coded up by the students using a rotation matrix for roll and yaw (combined into a single matrix in a method)
- The students are responsible for coding the transforms and for making sure they are applying the rotations in the correct order.

Open Questions:
What removing Eigen looks like. I think the student section needs a better defined scope and I don't want them to see Eigen template errors week one. So it will likely result in abstracting the multiplication and matrix building into a data type they know (vector) and then writing helpers that convert back and forth.

coordinate_transform

### DIFF
--- a/coordiante_transform/launch/coordinate_transform.launch.py
+++ b/coordiante_transform/launch/coordinate_transform.launch.py
@@ -19,6 +19,7 @@
 # THE SOFTWARE.
 
 from launch import LaunchDescription
+from launch.substitutions import LaunchConfiguration
 from launch_ros.actions import Node
 
 
@@ -27,6 +28,9 @@ def generate_launch_description():
         Node(
             package='coordinate_transform',
             executable='coordinate_transform_node',
-            output='screen'
+            output='screen',
+            parameters=[
+                {'use_sim_time': LaunchConfiguration('use_sim_time', default='false')}
+            ]
         )
     ])
--- a/coordiante_transform/launch/coordinate_transform.launch.py
+++ b/coordiante_transform/launch/coordinate_transform.launch.py
@@ -1,0 +1,32 @@
+# Copyright 2021 RoboJackets
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from launch import LaunchDescription
+from launch_ros.actions import Node
+
+
+def generate_launch_description():
+    return LaunchDescription([
+        Node(
+            package='coordinate_transform',
+            executable='coordinate_transform_node',
+            output='screen'
+        )
+    ])
--- a/coordiante_transform/package.xml
+++ b/coordiante_transform/package.xml
@@ -13,6 +13,7 @@
     <depend>rclcpp_components</depend>
     <depend>tf2_ros</depend>
     <depend>tf2_eigen</depend>
+    <depend>visualization_msgs</depend>
     <depend>stsl_interfaces</depend>
 
     <test_depend>ament_lint_auto</test_depend>
--- a/coordiante_transform/package.xml
+++ b/coordiante_transform/package.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+    <name>coordinate_transform</name>
+    <version>0.0.0</version>
+    <description>TODO: Package description</description>
+    <maintainer email="jgibson37@gatech.edu">Jason Gibson</maintainer>
+    <license>MIT</license>
+
+    <buildtool_depend>ament_cmake</buildtool_depend>
+
+    <depend>rclcpp</depend>
+    <depend>rclcpp_components</depend>
+    <depend>tf2_ros</depend>
+    <depend>tf2_eigen</depend>
+    <depend>stsl_interfaces</depend>
+
+    <test_depend>ament_lint_auto</test_depend>
+    <test_depend>ament_lint_common</test_depend>
+
+    <export>
+        <build_type>ament_cmake</build_type>
+    </export>
+</package>
--- a/coordiante_transform/package.xml
+++ b/coordiante_transform/package.xml
@@ -3,7 +3,7 @@
 <package format="3">
     <name>coordinate_transform</name>
     <version>0.0.0</version>
-    <description>TODO: Package description</description>
+    <description>A package for transforming a fiducial marker from camera frame to body frame</description>
     <maintainer email="jgibson37@gatech.edu">Jason Gibson</maintainer>
     <license>MIT</license>
 
--- a/coordiante_transform/src/coordinate_transform_component.cpp
+++ b/coordiante_transform/src/coordinate_transform_component.cpp
@@ -30,15 +30,18 @@
 
 using namespace std::chrono_literals;
 
-namespace coordinate_transform {
+namespace coordinate_transform
+{
 class CoordinateTransformComponent : public rclcpp::Node
 {
 public:
   explicit CoordinateTransformComponent(const rclcpp::NodeOptions & options)
-  : rclcpp::Node("coordinate_transformer", options), tf_buffer_(get_clock()), tf_listener_(tf_buffer_)
+  : rclcpp::Node("coordinate_transformer", options), tf_buffer_(get_clock()), tf_listener_(
+      tf_buffer_)
   {
-    tag_sub_ = create_subscription<stsl_interfaces::msg::TagArray>("/aruco_tag_detector/tags",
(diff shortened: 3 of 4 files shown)


JasonGibson274:
Check error message for dimensional issues. Leave in the Eigen code if we can get around it.

coordiante_transform/launch/coordinate_transform.launch.py
```diff
@@ -0,0 +1,32 @@
...
+from launch import LaunchDescription
+from launch_ros.actions import Node
+
+
+def generate_launch_description():
+    return LaunchDescription([
+        Node(
+            package='coordinate_transform',
+            executable='coordinate_transform_node',
+            output='screen'
```
barulicm: This node will probably need to know about simulation time to play nicely with TF.

```suggestion
            output='screen',
            parameters=[
                {'use_sim_time': LaunchConfiguration('use_sim_time', default='false')}
            ]
```

coordiante_transform/package.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+    <name>coordinate_transform</name>
+    <version>0.0.0</version>
+    <description>TODO: Package description</description>
```
barulicm:
```suggestion
    <description>A package for experimenting with coordinate frame transformations.</description>
```

coordiante_transform/src/coordinate_transform_component.cpp
```diff
@@ -0,0 +1,187 @@
...
+using namespace std::chrono_literals;
+
+namespace coordinate_transform {
+class CoordinateTransformComponent : public rclcpp::Node
+{
+public:
+  explicit CoordinateTransformComponent(const rclcpp::NodeOptions & options)
+  : rclcpp::Node("coordinate_transformer", options), tf_buffer_(get_clock()), tf_listener_(tf_buffer_)
+  {
+    tag_sub_ = this->create_subscription<stsl_interfaces::msg::TagArray>("/aruco_tag_detector/tags",
```
barulicm: I'm personally not a fan of the ROS tutorial code's overuse of `this->`. Open to discussion on this one, but I'm in favor of removing this everywhere.

```suggestion
    tag_sub_ = create_subscription<stsl_interfaces::msg::TagArray>("/aruco_tag_detector/tags",
```

coordiante_transform/src/coordinate_transform_component.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    tag_publisher_ = this->create_publisher<stsl_interfaces::msg::TagArray>("~/tags_transformed", 1);
+    visualization_pub_ = this->create_publisher<visualization_msgs::msg::MarkerArray>("~/tags_visual", 1);
+  }
+
+private:
+  tf2_ros::Buffer tf_buffer_;
+  tf2_ros::TransformListener tf_listener_;
+  rclcpp::Subscription<stsl_interfaces::msg::TagArray>::SharedPtr tag_sub_;
+  rclcpp::Publisher<stsl_interfaces::msg::TagArray>::SharedPtr tag_publisher_;
```
barulicm: nit: We should stay consistent with when we do and don't abbreviate words. 'pub'/'sub' everywhere or 'publisher'/'subscriber' everywhere.

```suggestion
  rclcpp::Publisher<stsl_interfaces::msg::TagArray>::SharedPtr tag_pub_;
```

coordiante_transform/src/coordinate_transform_component.cpp
```diff
@@ -0,0 +1,187 @@
...
+    const Eigen::Isometry3d eigen_transform = tf2::transformToEigen(tf_transform);
+
+    // creates a matrix that goes from camera to standard ROS coordinates
+    Eigen::Matrix4d optical = getRotMatForOpticalFrame();
+
+    // BEGIN STUDENT CODE
+    // create a new tag array message
+    stsl_interfaces::msg::TagArray new_tag_array_msg;
+    // copy the header information
+    new_tag_array_msg.header = tag_array_msg->header;
```
barulicm: Header is just frame and stamp now (no more seq), so I'd suggest it may be clearer to just copy the stamp and set the frame_id. (Rather than copying the whole header and re-writing the frame_id).

```suggestion
    new_tag_array_msg.header.stamp = tag_array_msg->header.stamp;
```

coordiante_transform/src/coordinate_transform_component.cpp
```diff
@@ -0,0 +1,187 @@
...
+      new_tag.pose.position.y = pose(1);
+      new_tag.pose.position.z = pose(2);
+
+      // get the orientation of the tag
+      Eigen::Matrix4d tag_orientation = convertQuatToRotMat(tag_array_msg->tags[i].pose.orientation);
+      tag_orientation = eigen_transform.matrix() * optical * tag_orientation;
+      new_tag.pose.orientation = convertRotMatToQuat(tag_orientation);
+
+      new_tag_array_msg.tags.push_back(new_tag);
+    }
```
barulicm: Prefer range-based for loops

```suggestion
    for(const auto& old_tag : tag_array_msg->tags) {
      stsl_interfaces::msg::Tag new_tag;
      new_tag.id = old_tag.id;

      geometry_msgs::msg::Point old_tag_position = old_tag.pose.position;

      // extract the important parts of the pose
      Eigen::Vector4d pose = Eigen::Vector4d(old_tag_position.x, old_tag_position.y, old_tag_position.z, 1);

      // apply the transform
      pose = eigen_transform.matrix() * optical * pose;

      new_tag.pose.position.x = pose(0);
      new_tag.pose.position.y = pose(1);
      new_tag.pose.position.z = pose(2);

      // get the orientation of the tag
      Eigen::Matrix4d tag_orientation = convertQuatToRotMat(old_tag.pose.orientation);
      tag_orientation = eigen_transform.matrix() * optical * tag_orientation;
      new_tag.pose.orientation = convertRotMatToQuat(tag_orientation);

      new_tag_array_msg.tags.push_back(new_tag);
    }
```

coordiante_transform/src/coordinate_transform_component.cpp
```diff
@@ -0,0 +1,187 @@
...
+    tag_publisher_->publish(new_tag_array_msg);
+    // END STUDENT CODE
+
+    // visualization code
+    visualization_msgs::msg::MarkerArray marker_array;
+    for(stsl_interfaces::msg::Tag tag : new_tag_array_msg.tags)
```
barulicm: Use const references in read-only range-based for loops
```suggestion
    for(const stsl_interfaces::msg::Tag& tag : new_tag_array_msg.tags)
```

coordiante_transform/src/coordinate_transform_component.cpp
```diff
@@ -0,0 +1,187 @@
...
+    {
+      visualization_msgs::msg::Marker marker;
+      marker.header.frame_id = "base_footprint";
+      marker.header.stamp = tag_array_msg->header.stamp;
+      marker.id = tag.id;
+      marker.type = visualization_msgs::msg::Marker::CUBE;
+      marker.action = visualization_msgs::msg::Marker::ADD;
+      marker.lifetime = rclcpp::Duration(0.25s);
```
barulicm: For discussion: Should we just use the seconds+nanoseconds constructor instead and avoid having to import the literals namespace above?
```suggestion
      marker.lifetime = rclcpp::Duration(0, 250000000);
```

coordiante_transform/src/coordinate_transform_component.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+      marker.pose.orientation.x = tag.pose.orientation.x;
+      marker.pose.orientation.y = tag.pose.orientation.y;
+      marker.pose.orientation.z = tag.pose.orientation.z;
+      marker.pose.orientation.w = tag.pose.orientation.w;
+
+      marker.pose.position.x = tag.pose.position.x;
+      marker.pose.position.y = tag.pose.position.y;
+      marker.pose.position.z = tag.pose.position.z;
```
barulicm: You should be able to just copy the whole pose object, instead of each individual member.

```suggestion
      marker.pose = tag.pose;
```

coordiante_transform/src/coordinate_transform_component.cpp
```diff
@@ -0,0 +1,187 @@
...
+    R_yaw << cos(M_PI/2), sin(M_PI/2), 0, 0,
+            -sin(M_PI/2), cos(M_PI/2), 0, 0,
+            0, 0, 1, 0,
+            0,0,0,1;
+    return R_yaw * R_roll;
+    // END STUDENT CODE
+  }
+
+  Eigen::Matrix4d convertQuatToRotMat(geometry_msgs::msg::Quaternion quat) {
+    Eigen::Quaterniond e_quat = Eigen::Quaterniond(quat.w, quat.x, quat.y, quat.z);
```
barulicm: Potential alternative. I don't feel strongly about this one. It is kinda nice to not have to worry about getting the ordering of the members correct.

```suggestion
    Eigen::Quaterniond e_quat;
    tf2::fromMsg(quat, e_quat);
```

coordiante_transform/src/coordinate_transform_component.cpp
```diff
@@ -0,0 +1,187 @@
...
+    Eigen::Matrix4d result_mat = Eigen::Matrix4d::Zero();
+    result_mat.block<3,3>(0,0) = e_quat.normalized().toRotationMatrix();
+    result_mat(3,3) = 1.0;
```
barulicm: Potential alternative:
```suggestion
    Eigen::Matrix4d result_mat = Eigen::Matrix4d::Identity();
    result_mat.block<3,3>(0,0) = e_quat.normalized().toRotationMatrix();
```

coordiante_transform/src/coordinate_transform_component.cpp
```diff
@@ -0,0 +1,187 @@
...
+  geometry_msgs::msg::Quaternion convertRotMatToQuat(Eigen::Matrix4d homo_mat) {
+    geometry_msgs::msg::Quaternion quat;
+    Eigen::Quaterniond e_quat = Eigen::Quaterniond(homo_mat.block<3,3>(0,0));
+
+    quat.x = e_quat.x();
+    quat.y = e_quat.y();
+    quat.z = e_quat.z();
+    quat.w = e_quat.w();
+
+    return quat;
```
barulicm: Can be simplified.

```suggestion
    return tf2::toMsg(Eigen::Quaterniond(homo_mat.block<3,3>(0,0));
```

barulicm (commented):
Looks pretty good! A handful of suggested edits. Nothing major.

coordiante_transform/src/coordinate_transform_component.cpp
JasonGibson274: I agree with this. Especially so week 1 we don't have to explain this as much.

coordiante_transform/src/coordinate_transform_component.cpp
JasonGibson274: I think for week one yes, but can re evaluate later on.

barulicm (approved):

PR #184, opened 2021-06-24
Adds peak_finder package for optimization project

barulicm:
This PR...

- Adds the peak_finder package & node
  The peak_finder node provides the /park_at_peak action that encapsulates the optimization project
- Adds optimization project launch file
  This launch file serves as the entry point for the optimization project. The idea is the students will use this launch file to bring everything up and then can trigger the behavior by sending a goal to the /park_at_peak action from a second terminal.
  This launch file also brings up rviz with a display layout that includes the elevation map visualization.
- Reconfigures nav stack param file to use new custom plugins from STSL
- Style fixes to get everything passing the linting tests again

Demo video:

Peek 2021-06-22 23-35

PR #187, opened 2021-07-11
A* Planner Plugin

barulicm:
This PR is blocked by #184. The diff seems to be a bit buggy (showing changes form previously merged PRs). This will hopefully go away when that PR is merged.

This PR adds a new `nav2_common::GlobalPlanner` plugin implementing the A* path planner. A new parameters file for the nav stack is added that configures the A* planner plugin.


astar_path_planner/package.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>astar_path_planner</name>
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
```
JasonGibson274: TODO

astar_path_planner/package.xml
```diff
@@ -0,0 +1,23 @@
...
+  <maintainer email="matthew.barulic@gmail.com">Matthew Barulic</maintainer>
+  <license>TODO: License declaration</license>
```
JasonGibson274: TODO

astar_path_planner/src/astar_path_planner.cpp
```diff
@@ -0,0 +1,186 @@
...
+      const Point neighbor{point.x() + dx, point.y() + dy};
+      if (!PointInCollision(neighbor)) {
+        neighbors.push_back(neighbor);
+      }
+    }
+  }
+  return neighbors;
+}
+
+double AStarPathPlanner::GetPathCost(const std::vector<Point> & path)
```
JasonGibson274: This doesn't have to be calculated from scratch each time. You can use the previous cost once you remove the heuristic and only do a single step.

astar_path_planner/src/astar_path_planner.cpp
```diff
@@ -0,0 +1,186 @@
...
+  costmap->convexFillCells(footprint_in_map, map_cells);
+
+  return std::any_of(
+    map_cells.begin(), map_cells.end(), [costmap](const auto & cell) {
+      const auto cost = costmap->getCost(cell.x, cell.y);
+      return cost == 253;
+    });
+}
+
+template<typename InputIterator>
```
JasonGibson274: Why is this templated? It is always a Point iterator right?

astar_path_planner/src/astar_path_planner.cpp
```diff
@@ -0,0 +1,186 @@
...
+    frontier.pop();
+    const auto last_state = path.back();
+
+    if (expanded.find(last_state) == expanded.end()) {
+      expanded.insert(last_state);
+
+      if (IsGoal(last_state)) {
+        std::vector<Point> reduced_path{path};
+        ReducePath(reduced_path);
+        return reduced_path;
```
JasonGibson274: I think we should return both the reduced path and the raw path for visualization purposes.

astar_path_planner/src/astar_path_planner_plugin.cpp
```diff
@@ -0,0 +1,100 @@
...
+  void activate() override {}
+
+  void deactivate() override {}
+
+  nav_msgs::msg::Path createPlan(
+    const geometry_msgs::msg::PoseStamped & start,
+    const geometry_msgs::msg::PoseStamped & goal) override
+  {
+    nav_msgs::msg::Path path;
+    path.header.stamp = node_->now();
```
JasonGibson274, line 65: Is there a way to make this time match the pose time? I'm not sure if start is the current position or not.

JasonGibson274 (commented):
Overall I think this is good. I just had a couple smaller questions about some C++ specific and then a potential simpler calculating total cost.

astar_path_planner/src/astar_path_planner.cpp
barulicm: Yeah, it'll always be a `std::vector<Point>::iterator` in this case. This is mostly written this way out of habit, because this is how you're supposed to write container-agnostic algorithms.

astar_path_planner/src/astar_path_planner_plugin.cpp
barulicm: Start is _usually_ the current pose (though that's really up to the caller of the service).

My understanding is that the path header stamp is expected to represent when the path was generated, not necessarily tied to the timestamps of the input poses.

astar_path_planner/src/astar_path_planner.cpp
barulicm: hhmmmmmmmm.... that's probably a good idea.
My instinct then is to separate the reduce path function out of this class entirely and the next layer up can explicitly call it and have access to both.

astar_path_planner/src/astar_path_planner.cpp
barulicm: True. I can switch it. My only hesitation is whether that's trickier to explain to the students.

JasonGibson274: I'll file this under showing them what they should be doing rather than what is immediately easy to understand. I wouldn't want to see this in team code so I think we should not show it.

JasonGibson274 (approved):
LGTM

PR #189, opened 2021-07-24
Kalman Filter code

barulicm:
This PR adds the mineral_deposit_tracking package which includes the kalman filter project node.

### DIFF
--- a/mineral_deposit_tracking/package.xml
+++ b/mineral_deposit_tracking/package.xml
@@ -3,7 +3,7 @@
 <package format="3">
   <name>mineral_deposit_tracking</name>
   <version>0.0.0</version>
-  <description>TODO: Package description</description>
+  <description>Implements a Kalman Filter for tracking mineral deposits</description>
   <maintainer email="matthew.barulic@gmail.com">Matthew Barulic</maintainer>
   <license>MIT</license>
 
--- a/mineral_deposit_tracking/package.xml
+++ b/mineral_deposit_tracking/package.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>mineral_deposit_tracking</name>
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
+  <maintainer email="matthew.barulic@gmail.com">Matthew Barulic</maintainer>
+  <license>MIT</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <depend>rclcpp</depend>
+  <depend>rclcpp_components</depend>
+  <depend>stsl_interfaces</depend>
+  <depend>geometry_msgs</depend>
+  <depend>eigen3_cmake_module</depend>
+  <depend>eigen</depend>
+  <depend>tf2_ros</depend>
+  <depend>tf2_eigen</depend>
+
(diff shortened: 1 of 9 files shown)


mineral_deposit_tracking/package.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>mineral_deposit_tracking</name>
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
```
JasonGibson274: TODO

mineral_deposit_tracking/src/mineral_deposit_tracker.cpp
```diff
@@ -0,0 +1,124 @@
...
+      const Eigen::Vector2d measurement{map_frame_position.x(), map_frame_position.y()};
+
+      const Eigen::Matrix2d covariance = Eigen::Matrix2d::Identity() * 0.01;
+
+      filter_.MeasurementUpdate(measurement, covariance);
+
+      const auto estimated_position = filter_.GetEstimate();
+
+      geometry_msgs::msg::PoseStamped output_msg;
+      output_msg.header.stamp = now();
```
JasonGibson274, line 98: Should this be the timestamp of the message?

mineral_deposit_tracking/src/mineral_deposit_tracker.cpp
```diff
@@ -0,0 +1,124 @@
...
+        "map", msg->header.frame_id, msg->header.stamp, rclcpp::Duration::from_seconds(
+          0.1));
+
+      Eigen::Vector3d map_frame_position;
+
+      tf2::doTransform(sensor_frame_position, map_frame_position, transform);
+
+      const Eigen::Vector2d measurement{map_frame_position.x(), map_frame_position.y()};
+
+      const Eigen::Matrix2d covariance = Eigen::Matrix2d::Identity() * 0.01;
```
JasonGibson274, line 91: cov should come from the message

JasonGibson274 (commented):
LGTM

mineral_deposit_tracking/src/mineral_deposit_tracker.cpp
barulicm: I don't know if it matters much. I don't see a strong connection between a specific message and the published estimate, so I set to `now()`.

mineral_deposit_tracking/src/mineral_deposit_tracker.cpp
barulicm: Should is a strong word :)

If the covariance is in the message, should it reflect the covariance on the range/heading (as opposed to x/y)? And then should this node be doing math to convert it to x/y covariance for the filter?

JasonGibson274: I think cov should be in x,y since that is what the message says but then convert it back out. I'm okay with just leaving it as this since that sounds more complex than it is probably worth though.

barulicm: The message doesn't use x/y, it uses range/heading.

barulicm: We'll revisit this when we get around to converting it into an exercise. It's not till week 6, so we've got time.

JasonGibson274 (approved):

PR #192, opened 2021-08-15
Week 1 Exercise

barulicm:
This PR adds the instructions for the week 1 exercises and turns the code into starter code.

See the rendered version of the instructions here: https://github.com/RoboJackets/software-training/blob/dev/barulicm/week1_exercise/projects/week_1/Instructions.md

This also adds a script to generate tables of contents with the doctoc tool.


projects/week_1/Instructions.md
```diff
@@ -0,0 +1,332 @@
...
+  - [2.11 Modernize the for loop](#211-modernize-the-for-loop)
+  - [2.12 Commit your new code in git](#212-commit-your-new-code-in-git)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## 1 Background
+
+This week's robotics videos covered how coordinate frames give us information about the reference frame for a measurement, and the math used to transform data between frames. In this project, we're going to practice that math a bit by transforming measurements from our robot from one frame to another.
+
+Our robot uses a camera to track ArUco tag fiducials. We'll be using these detections in a later project to localize our robot. For now, we just want to take these detections, which are measured in the camera's optical frame, and transform them to the robot's "base_footprint" frame. This will give you a glimpse into the process used to track objects against different reference frames.
```
JasonGibson274: I'd include a fiducial as an example.

projects/week_1/Instructions.md
```diff
@@ -0,0 +1,332 @@
...
+  - [2.9 Build project](#29-build-project)
+  - [2.10 Run project](#210-run-project)
+  - [2.11 Modernize the for loop](#211-modernize-the-for-loop)
+  - [2.12 Commit your new code in git](#212-commit-your-new-code-in-git)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## 1 Background
+
+This week's robotics videos covered how coordinate frames give us information about the reference frame for a measurement, and the math used to transform data between frames. In this project, we're going to practice that math a bit by transforming measurements from our robot from one frame to another.
```
JasonGibson274: This paragraph confuses me, might I propose something like.

This weeks' robotics videos covered the basics of coordinate frames. We will be practicing the math covered in order to transform a measurement coming from an attached camera to the body frame of the robot.

projects/week_1/Instructions.md
JasonGibson274: They don't know what localize is at this point

projects/week_1/Instructions.md
```diff
@@ -0,0 +1,332 @@
...
+
+Locate the include block at the top of the file.
+
+**Tip** We'll mark locations in the starter code where you'll need to add code with `// BEGIN STUDENT CODE` and `// END STUDENT CODE`.
+
+In the marked student code section, add two lines to include the new headers. You'll need to include `<array>` and `<vector>`.
+
+With that done, we can now use those containers in this file.
+
+### 2.4 Implement the rotation matrix helper function
```
JasonGibson274: I still think we should split the naming for these matricies as 

rotation: 3x3
transformation: 4x4

JasonGibson274: Mostly since that is what I have in the video lecture

projects/week_1/Instructions.md
```diff
@@ -0,0 +1,332 @@
...
+</details>
+
+Next, let's create `Eigen::Matrix4d` objects using our data arrays:
+
+```c++
+Eigen::Matrix4d R_roll(R_roll_data.data());
+Eigen::Matrix4d R_yaw(R_yaw_data.data());
+```
+
+Finally, modify the return statement to return the result of `R_yaw` multiplied by `R_roll`.
```
JasonGibson274: I would add a hint here about what order the rotation matricies should be multiplied in.

R_roll * R_yaw

ROS should be yaw pitch roll

projects/week_1/Instructions.md
```diff
@@ -0,0 +1,332 @@
...
+
+Here's an example of what things might look like if something's not right. In this case, the rotation matrices weren't setup correctly.
+
+![Non-working code demo](non-working-demo.gif)
+
+### 2.11 Modernize the for loop
+
+Now that we've got our code working, let's take a moment to improve our implementation by changing our index-based for loop to a range-based for loop.
+
+Rewrite your for loop using the range-based syntax with the element variable named `old_tag`. Be sure to replace all instances of `tag_array_msg->tags[i]` with `old_tag`.
```
JasonGibson274, line 332: we should comment that i is the variable being incremented. They might not be familiar with that convention.

projects/week_1/Instructions.md
```diff
@@ -0,0 +1,332 @@
...
+```c++
+Eigen::Vector4d position(
+  tag_array_msg->tags[i].pose.position.x,
+  tag_array_msg->tags[i].pose.position.y,
+  tag_array_msg->tags[i].pose.position.z,
+  1
+);
+```
+
+Now, apply the transformations by multiplying the transformation objects, `camera_to_base_transform` and `camera_optical_to_conventional_transform`, with `position`. Assign the result of that multiplication back to `position`. Remember, the order of this multiplication matters. Be sure to apply `camera_optical_to_conventional_transform` before `camera_to_base_transform`.
```
JasonGibson274, line 230: I would put the written out multiplication order into a hint.

projects/week_1/Instructions.md
```diff
@@ -0,0 +1,332 @@
...
+
+### 2.8 Transform tag orientation
+
+To transform the tag's orientation, we're going to follow a similar process. We'll start by extracting the rotation from the old tag message. We've provided a function to help you with this. Declare a variable named `tag_orientation` of type `Eigen::Matrix4d` right after you position code. Initialize `tag_orientation` by calling `quaternionMessageToTransformationMatrix()`, passing it `tag_array_msg->tags[i].pose.orientation`.
+
+Now, apply both transformations to `tag_orientation`. This should look basically identical to how you applied the transformations to `position`.
+
+Finally, we need to convert our new orientation back into a quaternion message and set it in the `new_tag` message. Here again, we've got a helper function for you called `transformationMatrixToQuaternionMessage()`. Call this function, passing `tag_orientation`, and assign the result to `new_tag.pose.orientation`.
+
+And that's it! We've written all the code we need to make this node complete.
```
JasonGibson274: This section feel a little weird to me since you are using a 4x4 when you only need a 3x3. I get the code is simpler but I don't want to confuse them. We can talk about this tonight. 

Maybe something as simple as explaining the above point would be good enough for me. i.e. that the translation vector in the 4x4 block matrix doesn't matter here.

projects/week_1/Instructions.md
```diff
@@ -0,0 +1,332 @@
...
+  1 package had stderr output: coordinate_transform
+```
+
+If you do have any errors, you'll need to fix them before you can run your node. We recommend starting with the first error reported, reading the error message carefully, fixing it, and building again. Repeat that process until all errors are resolved. Of course, if you're not sure how to fix any given error, reach out to the trainers.
+
+### 2.10 Run project
+
+To run this project, we'll need two terminal windows or tabs.
+
+In the first terminal, after sourcing the underlay setup file and the training workspace overlay setup file, launch the week 1 launch file.
```
JasonGibson274: I would split the sourcing overlay and underlay into a separate box. I don't want them to start to think those three commands equal a launch, which is what I would think reading this.

projects/week_1/Instructions.md
```diff
@@ -0,0 +1,332 @@
...
+
+Or, for keyboard control:
+
+```bash
+$ ros2 run teleop_twist_keyboard teleop_twist_keyboard
+```
+
+Now you should be able to drive the robot around and see the tag detections visualized in the rviz window.
+
+If everything's working correctly, you'll see something like this:
```
JasonGibson274: Where the blue box is the location rviz thinks the tag should be and the green one is where your code placed it.

projects/week_1/Instructions.md
barulicm: I'll need to update `getRotationMatrixForOpticalFrame()` as well.

projects/week_1/Instructions.md
JasonGibson274: I think something simple like saying the rotation is represented by the rotation matrix in the transformation matrix.

projects/week_1/Instructions.md
```diff
@@ -0,0 +1,350 @@
...
+```
+
+Now, apply the transformations by multiplying the transformation objects, `camera_to_base_transform` and `camera_optical_to_conventional_transform`, with `position`. Assign the result of that multiplication back to `position`. Remember, the order of this multiplication matters. Be sure to apply `camera_optical_to_conventional_transform` before `camera_to_base_transform`.
+
+<details>
+<summary><b>Hint:</b> Ordering transformations</summary>
+<p>Transforms that are applied first should appear on the right of a multiplication chain. To apply A, then B, then C, we would write <code>C * B * A</code>.</p>
+</details>
+
+The last thing we need to do for this section is to copy the new position's coordinates into the `new_tag` message. Copying the x value looks like this:
```
JasonGibson274, line 237: Should this be up where we first discuss getTransformationMatrixForOpticalFrame()?

JasonGibson274 (approved):
LGTM just with the one question or ordering.

projects/week_1/Instructions.md
barulicm: It can be if we want. Right now, the optical frame instructions just tell them the order to use.

JasonGibson274: It's fine either way

PR #193, opened 2021-08-17
Syllabus

barulicm:
Adds the syllabus document we discussed at our meeting.

### DIFF
--- a/Syllabus.md
+++ b/Syllabus.md
@@ -1,6 +1,6 @@
 # RoboJackets Software Training - Fall 2021
 
-TODO
+The RoboJackets software training program is designed to prepare you to work on software for the RoboJackets competitive teams. This program offers practical experience writing code using C++ and ROS. It also covers some of the fundamental concepts of robotics used by all of our teams.
 
 ## Trainers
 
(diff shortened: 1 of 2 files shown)


Syllabus.md
```diff
@@ -0,0 +1,80 @@
...
+  The GitHub repository that hosts most of the resources for the software training program, including project instructions and starter code.
+
+- [STSL Repository](https://github.com/RoboJackets/stsl)
+
+  The Software Training Support Library repository. This holds support code for the training projects. Open issues against this repository if you find bugs with the robots or simulator.
+
+## Prerequisites
+
+We will assume that students are familiar with the concepts covered in [AP Computer Science A](https://apstudents.collegeboard.org/courses/ap-computer-science-a). Topics we assume knowledge of will be briefly covered in Week 0 content. Topics covered in AP Computer Science A that have special syntax, properties, or behavior in C++ will be covered in the main course content.
+
```
JasonGibson274, line 50: Add a section about the pre-reps for the robotics track

"Students should be familiar with AB level calculus. Namely the concepts of a derivatives and integrals."

JasonGibson274 (approved):
Overall questions

1. Do we belong anywhere on this syllabus?
2. Can I clear enrollment and all questions on Piazza?

Otherwise LGTM

barulicm:
1. Good question. I'm not sure if we belong on the syllabus at all. My instinct is to leave us off, but I don't have any reasons one way or the other.
2. We can. We also don't have to use that exact piazza workspace. I don't know how Piazza's supposed to work when it comes to cleaning out an old class or making a new one. I just linked to one I had access to.

It looks like we have at least two right now
Fall 2020: https://piazza.com/class/kc4upgfmw3h199
Fall 2017: https://piazza.com/class/j4pry2dzzg76pl

JasonGibson274:
1. us being left off it good for me
2. I guess we can just create a new one each year so we can go back and look at the questions we had. Also easier than cleaning it IMO. We just need to export a list of all emails.

PR #195, opened 2021-08-30
Project 2

barulicm:
This PR
- Adds week 2 project instructions
- Adds week 2 project reference solution
- Starter code-ifies the obstacle_detector package

Rendered view of instructions: https://github.com/RoboJackets/software-training/blob/dev/barulicm/project2/projects/week_2/Instructions.md


projects/week_2/Instructions.md
```diff
@@ -0,0 +1,275 @@
...
+Now declare a variable to hold the HSV version of our input image. This should be a `cv::Mat` variable called `input_hsv`. Then, call the `cvtColor` function from OpenCV to do the conversion to HSV:
+
+```c++
+cv::Mat input_hsv;
+cv::cvtColor(input, input_hsv, cv::COLOR_BGR2HSV);
+```
+
+Note that the conversion type uses "BGR". This color space uses the same channels as RGB, just in the opposite order in the image data. OpenCV uses BGR as the default color space for color images.
+
+Next, declare another `cv::Mat` variable that will hold our output image, named `outupt`. We'll initialise this variable to be the same size as our input image while holding one channel of unsigned, 8-bit values.
```
VAM7686: typos
```suggestion
Next, declare another `cv::Mat` variable that will hold our output image, named `output`. We'll initialize this variable to be the same size as our input image while holding one channel of unsigned, 8-bit values.
```

projects/week_2/Instructions.md
```diff
@@ -0,0 +1,275 @@
...
+output.at<uint8_t>(dest_point) = 127;
+```
+
+Finally, after the end of both nested loops, return `output`.
+
+### 3.5 Call our functions in `ObstacleDetector`
+
+Both of our key functions are now implemented, so it's time to use them by calling them within `ObstacleDetector`. In [obstacle_detector.cpp](../../obstacle_detector/src/obstacle_detector.cpp), find the student code comment block at the end of the existing set of `#include` lines. Add an `#include` line for your header, "student_functions.hpp".
+
+Nest, find the student code comments that include `// Call FindColors()`. Within that comment block, you'll see a declared but uninitialized variable named `detected_colors`. Add an initializer for this variable that calls `FindColors`. The input image for `FindColors` is `cv_image->image`, and the color range is given by `min_color` and `max_color`.
```
VAM7686: typo.
```suggestion
Next, find the student code comments that include `// Call FindColors()`. Within that comment block, you'll see a declared but uninitialized variable named `detected_colors`. Add an initializer for this variable that calls `FindColors`. The input image for `FindColors` is `cv_image->image`, and the color range is given by `min_color` and `max_color`.
```

projects/week_2/Instructions.md
```diff
@@ -0,0 +1,275 @@
...
+
+Let's start by declaring our output image variable. We'll call it `output` and use `cv::Mat` as its type. Initialize it by passing `map_size` and `CV_8UC1` to its constructor. `map_size` just holds the desired height and width for our output image. `CV_8UC1` tells OpenCV that this image will have one channel of unsigned, 8-bit values. (Our color images so far have been using `CV_8UC3` to store three channels of unsigned, 8-bit values).
+
+Next, setup two nested loops to iterate over all pixel positions in `output`. The first loop should iterate from `y = 0` to `y = output.rows-1`. The second loop should iterator from `x = 0` to `x = output.cols-1`.
+
+In the body of our inner nested loop, we're going do these steps:
+
+1. Create a homogeneous point for our current output position
```
VAM7686, line 193: Maybe add a link explaining why homogeneous coordinates are used?

VAM7686 (changes requested):
Looks great! Just some typos in the instructions.

projects/week_2/Instructions.md
barulicm: @JasonGibson274 iirc we're explaining homogeneous coordinates in the videos, right?

barulicm:
Noticed I forgot the git pull / commit steps at the beginning and end of the instructions. Need to add those.

projects/week_2/Instructions.md
JasonGibson274: In week 1 we mention that homogenous coordinates augments the vector with a 1. The recitation guide will have a better explanation with why that's a reasonable thing to do in week 1. So I think something as simple as a reminder to just add a one to the end of the vector should cover this.

I'll likely have an exercise in week 2 for further practice.

projects/week_2/Instructions.md
```diff
@@ -0,0 +1,275 @@
...
+
+<details>
+<summary><b>Hint:</b> How to include a header</summary>
+<p>To include a header file named "my_header.hpp", we'd write the code below. Note that the use of quotes ("") instead of angle brackets (<>) indicates that this header is part of our current package.</p>
+<code>#include "my_header.hpp"</code>
+</details>
+
+In our new header file (student_functions.hpp), we need to add header guards. Add the three lines that define the header guards for this file. Remember that header guards look like this:
+
+```c++
```
JasonGibson274, line 95: We should check with teams. IDK if they are using this or #pragma once

projects/week_2/Instructions.md
```diff
@@ -0,0 +1,275 @@
...
+#define STUDENT_FUNCTIONS_HPP  // Define the unique macro
+
+#endif  // Close the pre-processor if block
+```
+
+Finally, to include our new files into our package's build rules, we need to edit obstacle_detector's [CMakeLists.txt](../../obstacle_detector/CMakeLists.txt).
+
+Find the student code block in the call to `add_library`. Add our new implementation file, `src/student_code.cpp` to this function call.
+
+**Note:** We don't need to add our header file, because it gets compiled as part of the files that include it. Most C++ projects omit header files from the build rules unless there are special technical reasons that require them to be there.
```
JasonGibson274: I can go either way, but is the second sentence ever going to come up in training? It feels like extra information that may not be important.

projects/week_2/Instructions.md
```diff
@@ -0,0 +1,275 @@
...
+Now declare a variable to hold the HSV version of our input image. This should be a `cv::Mat` variable called `input_hsv`. Then, call the `cvtColor` function from OpenCV to do the conversion to HSV:
+
+```c++
+cv::Mat input_hsv;
+cv::cvtColor(input, input_hsv, cv::COLOR_BGR2HSV);
+```
+
+Note that the conversion type uses "BGR". This color space uses the same channels as RGB, just in the opposite order in the image data. OpenCV uses BGR as the default color space for color images.
+
+Next, declare another `cv::Mat` variable that will hold our output image, named `output`. We'll initialize this variable to be the same size as our input image while holding one channel of unsigned, 8-bit values.
```
JasonGibson274: Do we talk about what an unsigned bit is in the C++ videos? I know we don't in the robotics side.

projects/week_2/Instructions.md
```diff
@@ -0,0 +1,275 @@
...
+1. A constant `cv::Scalar` named `range_min`
+1. A constant `cv::Scalar` named `range_max`
+
+Finally, add the declaration for our second function. This function should be named `ReprojectToGroundPlane`. Its return type should be `cv::Mat`, and it should take these parameters:
+
+1. A constant `cv::Mat` named `input`
+1. A constant `cv::Mat` named `homography`
+1. A constant `cv::Size` named `map_size`
+
+### 3.3 Implement `FindColors`
```
JasonGibson274: If there are some bolded headings here that would be nice. This section is pretty long and hard to parse.

projects/week_2/Instructions.md
```diff
@@ -0,0 +1,275 @@
...
+
+Finally, after the end of the nested loops, return our `output` image.
+
+### 3.4 Implement `ReprojectToGroundPlane`
+
+`ReprojectToGroundPlane` warps `input` using the homography defined by the `homography` matrix. This function will create an output image with the size given by `map_size`. It will then iterate over each pixel position in the output image and calculate the corresponding pixel position in the input image with the homography matrix. Finally, it copies the value from the source position in the source image to the destination position in the output image. Any pixels that get mapped outside of the bounds of the source image will be set to 127 in the output image.
+
+Let's start by declaring our output image variable. We'll call it `output` and use `cv::Mat` as its type. Initialize it by passing `map_size` and `CV_8UC1` to its constructor. `map_size` just holds the desired height and width for our output image. `CV_8UC1` tells OpenCV that this image will have one channel of unsigned, 8-bit values. (Our color images so far have been using `CV_8UC3` to store three channels of unsigned, 8-bit values).
```
JasonGibson274: bits are explained better here then when they are first mentioned.

projects/week_2/Instructions.md
barulicm: The teams use a mix. I can add `#pragma once` to the recitation guide. All of the training code and videos sticks to guards.

projects/week_2/Instructions.md
barulicm: Yes, they get mentioned in the C++ videos

projects/week_2/Instructions.md
barulicm: Doesn't come up in training, but does come up in some team code when they're using Qt.

projects/week_2/Instructions.md
JasonGibson274: We can leave that to the teams then

projects/week_2/Instructions.md
JasonGibson274: Should we just remove it then?

barulicm: Removed.

projects/week_2/Instructions.md
barulicm: Moved this explanation up to the first use of `CV_8UC1`.

projects/week_2/Instructions.md
barulicm: Is [line 191](https://github.com/RoboJackets/software-training/pull/195/files#diff-40f8c92533a4fd9d22102627724f23c8350eb7a775f45bc2ebd48a2bfbffbd22R191) enough, or should it be more explicit?

> Start by declaring a constant `cv::Vec3d` named `dest_vec`. Initialize it by passing `x`, `y`, and `1` to the constructor.

projects/week_2/Instructions.md
barulicm: I'll break it into two sections. One will cover the color space background and conversion and the other will do the range checking.

barulicm: I ended up also moving the couple of paragraphs about color spaces up into the background section.

JasonGibson274 (approved):
LGTM

PR #199, opened 2021-09-20
VM credentials update

VAM7686:
Minor updates to the virtual machine setup instructions:
- New username and password credentials
- New prebuilt image download link


barulicm (approved):
Thanks!

PR #202, opened 2021-09-29
Change student_code to student_functions

thecharlesjenkins:
I believe this was a typo, this file is referred to as student_functions everywhere else.


VAM7686 (approved):
Yep just noticed that too

PR #203, opened 2021-09-29
Get rwm qos profile

thecharlesjenkins:
Should `rclcpp::SensorDataQoS()` be `rclcpp::SensorDataQoS().get_rmw_qos_profile()`? `create_camera_subscription` seems to want a `rmw_qos_profile_t` struct instead of a QOS object.


VAM7686 (approved):
I got the same problem. lgtm.

barulicm (commented):
Well, this is a little frustrating because I feel like this worked before, but it definitely doesn't.

Could you update the reference solution as well?

PR #206, opened 2021-10-14
Project 5

barulicm:
This PR adds starter code, instructions, and reference solution for Project 5: Mapping.

This also includes a few fixes to the obstacle detector project code. That node wasn't actually doing what it should've when handling axes and putting the occupancy grid message together. This may introduce merge conflicts with student's existing solution code, but I'm hoping they'll be minimal since there shouldn't be overlapping changes.

A couple of TODOs to insert more detailed math explanations if we have time.

working_fake_localizer


barulicm:
Thanks for those updates!

PR #210, opened 2021-11-10
Dev/control

JasonGibson274:
Adds the LQR tracking controller code and project 7 that implements it

working_lqr


barulicm (approved):
Good stuff. Beautifully smooth controller.
I did a pretty deep editing pass on the instructions and tweaked some smaller stuff in the code/

I also removed support for the particle filter localizer from the launch file. It doesn't actually work since it requires a steady feed of cmd_vel messages, which we don't have in this example. Definitely something we'll have to revisit when we come back to that node next year.

PR #211, opened 2021-11-12
Project 8

barulicm:
Adds project 8, which covers implementing an A* path planning plugin for Nav2.

The project also includes some practice with rosbag.


JasonGibson274 (approved):
LGTM

PR #213, opened 2022-06-29
Updates to project 1

barulicm:
- Fixes a bug in the launch file
- Fixes mistakes in the instructions
- Adds more explanation text to the instructions
- Removes the whole "write a traditional for loop and then upgrade it to range-based" bit. Felt it didn't flow well and added unnecessary confusion to the exercise. It also didn't really follow the "don't teach what you don't want them to write" principle.

__Note:__ This was tested with the changes in https://github.com/RoboJackets/stsl/pull/30


anshulm07 (approved):
Looks good to me!

idale0001 (approved):
Everything worked on my laptop and there aren't any issues I found reading through the instructions again. There are a few things I can see being a little unclear for new members with the current wording but this needs to be merged so its good for now

PR #214, opened 2022-07-02
Updates to Project 2

barulicm:
- Fixes a bug in the launch file
- Fixes mistakes in the instructions
- Adds more code hints throughout the instructions

**Note:** This was tested with the changes in https://github.com/RoboJackets/stsl/pull/30

### DIFF
--- a/projects/week_2/Instructions.md
+++ b/projects/week_2/Instructions.md
@@ -139,7 +139,7 @@ Finally, add the declaration for our second function. This function should be na
 1. A constant `cv::Size` named `map_size`
 
 ```c++
-cv::Mat ProjectToGroundPlane(const cv::Mat input,
+cv::Mat ReprojectToGroundPlane(const cv::Mat input,
                              const cv::Mat homography,
                              const cv::Size map_size);
 ```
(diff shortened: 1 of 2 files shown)


anshulm07 (approved):
Looks good to me!

matthewhannay567 (changes requested):
capture
The output I got is different than the example image in Section 2.
Otherwise, it looks fine.

barulicm:
@matthewhannay567 That looks like you're detecting blue instead of red. Are you sure you used BGR2HSV and not RGB2HSV?

matthewhannay567:
capture
@barulicm  This is my FindColors implementation.
Also I'm confused by the fact that the instructions say to use brackets with cv::Scalar, is that actually needed?

barulicm:
Ah ok, I see what's happening now.

The results you're seeing actually highlight most of the world around you as obstacles, and you're only seeing the blue tags as free space. This is because you're only comparing the hue (the first channel) and ignoring the other two channels (saturation and value). The color comparison if statement is meant to check all three channels (as shown in the reference solution):

```C++
if (input_color[0] >= range_min[0] && input_color[0] <= range_max[0] &&
  input_color[1] >= range_min[1] && input_color[1] <= range_max[1] &&
  input_color[2] >= range_min[2] && input_color[2] <= range_max[2])
{
  output.at<uint8_t>(r, c) = 255;
} else {
  output.at<uint8_t>(r, c) = 0;
}
```

The brackets (assuming you're talking about the `[0]` part) are required because the OpenCV types used here don't have comparison operators defined on them. We can't just write `input_color > range_min` because that operator doesn't actually exist. So, instead, we grab each element from the `Scalar`s and compare them individually.

Let me know if you have any ideas on how to make that section of the instructions more clear.

PR #215, opened 2022-07-23
Updates to Project 4

barulicm:
This PR makes some improvements and fixes some bugs in Project 4. The project code itself didn't actually need much work, but the navigation stack had a lot of lurking errors in it. Most of this PR ends up actually making nav fixes to get Project 4 working again.

Navigation related changes include:
- param_file -> nav_params_file to avoid name collision with gazebo's params_file argument
- nav2_recoveries -> nav2_behaviors
- Adds empty map server so projects that don't use real-time mapping can still navigate (disabling the static layer doesn't seem to work anymore)
- Updates navigation BT param names to match Humble changes


idale0001 (approved):
The project works on my machine and the changes to the instructions I think were a big help. I don't see anything that needs additional changes, should be ready to merge

PR #225, opened 2022-11-07
Project 7 Updates

JasonGibson274:
Corrects project 7 to remove some of the confusing parts about running LQR. Includes major changes to the task

1. The path to track was modified to show a better difference between LQR and PID. It tracks in reverse as well.
2. PID controller plugin was added in addition to LQR and is switched out by changing what plugin is loaded.
3. Fixed some formatting

I am still going to go through and making sure the instructions are good, but the code should be good.


setup_instructons/bare_metal_ubuntu_arm64.md
```diff
@@ -5,8 +5,7 @@
 ## Contents
 
 - [1 Install ROS](#1-install-ros)
-- [2 Install STSL](#2-install-stsl)
```
barulicm, line 8: Was this removal intentional?

JasonGibson274: No, but the section doesn't exist in the markdown so it should be gone. I fixed the numbering now.

JasonGibson274:
Instructions should be good now.

barulicm (approved):

PR #237, opened 2025-01-29
Update docker.md to fix so that it gets the correct .sh file

Robert-Jia00129:
The url for .sh file wasn't correct.


MukilanKarthikeyan:
thank you! fixed it with the addition of a curl command as well!

Robert-Jia00129:
I'm spending the following weeks on this software training repo. Would you be available in the following weeks? Can I also message you on RoboJackets slack?

